PR #28, opened 2014-11-27
ENH: moved tauchen routine to mc_tools. updated tests

spencerlyon2:
open PR for travis to build


spencerlyon2:
Don't think this is needed.

PR #30, opened 2014-12-03
Add `solve_vf`, `solve_pf`, `solve_both` methods and an `AbstractModel` type

spencerlyon2:
This PR adds an `AbstractModel` type that each of the types within the models subpackage now subtype.

In addition to the new type, we have new a few new functions:

* `init_values{T <: AbstractModel)(m::T)`: given a model object, this function returns the initial values used in the solutions notebook 
* `solve_vf(m::AbstractModel, [init=init_values(m); kwargs...])`: this function calls `init = init_values(m)` to construct a starting guess for the value function, then passes `f(x) = bellman_operator(m, x)` through `compute_fixed_point(f, init; kwargs...)` to compute the solution to the problem.
  * for models that don't implement `init_values` or `bellman_operator`  (`AssetPrices`, `LucasTree`), this function will send an error saying that `init_values` and/or `bellman_operator` is not defined for this model type
  * for the `ConsumerProblem` type in `ifp.jl` we write a special version of `solve_vf` that uses the more efficient `coleman_operator` instead of `bellman_operator`
* `solve_pf(m::AbstractModel, [init=init_values(m); kwargs...])`: This function gets the final value function from `vf = solve_vf(m, init; kwargs...)`, then computes and returns `get_greedy(m, vf)`
* `solve_both(m::AbstractModel, [init; kwargs...])`: This function uses the two previous functions to return both the final value function and policy function.

I have included tests for all of these methods (just that each of the functions does or doesn't run for a given model type).

If anyone has comments please let me know -- otherwise I'll merge this soon.

### DIFF
--- a/examples/jv_test.jl
+++ b/examples/jv_test.jl
@@ -1,4 +1,5 @@
 using QuantEcon
+using QuantEcon.Models
 using PyPlot
 
 wp = JvWorker(grid_size=25)
--- a/src/Models.jl
+++ b/src/Models.jl
@@ -9,8 +9,46 @@ using Distributions
 using Optim: optimize
 using Grid: CoordInterpGrid, BCnan, BCnearest, InterpLinear
 
+abstract AbstractModel
+
+"""
+Generic function to solve model via value function iteration.
+
+For this function to work
+"""
+
+function solve_vf(m::AbstractModel; kwargs...)
+    # make sure init_values is defined. If not provide a decent error message
+    T_model = typeof(m)
+    if !method_exists(init_values, (T_model,))
+        e = "init_values(m::$(T_model) must be defined to use default solver"
+        error(e)
+    end
+    init = init_values(m)
+
+    # hand off to routine that takes initial values as argument
+    solve_vf(m, init; kwargs...)
+end
+
+function solve_vf(m::AbstractModel, init; kwargs...)
+    # make sure bellman_operator is defined
+
+    T_model = typeof(m)
+    T_init = typeof(init)
+    if !method_exists(bellman_operator, (T_model, typeof(init)))
+        e = "bellman_operator(m::$(T_model), x::$(T_init))"
+        e *= " must be defined to use default solver"
+        error(e)
+    end
+
+    # solve this thing!
+    f(x) = bellman_operator(m, x)
+    compute_fixed_point(f, init; kwargs...)
+end
+
 export
 # types
+    AbstractModel,
     AssetPrices,
     CareerWorkerProblem,
     ConsumerProblem,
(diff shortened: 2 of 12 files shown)


spencerlyon2:
I still like this idea, but won't merge unless others in the team think it is appropriate. 

Any thoughts?

ZacCranko:
I like ``AbstractModel`` a lot, and think it's very Julian, and it would definitely serve to make the user experience a lot more slick. I would even prefer more abstraction and just have ``solve()``, with  with maybe ``method=:value`` to pick value function iteration, defaulting to something sensible.

On the other hand for quite simple things, like the methods relating to Markov chains, I'd be in favour of having less type structure, since the abstraction doesn't really simplify the code or add to the user experience. (For common tasks like checking if the inputs are sensible, it would be simple enough to define macros, e.g., @valid_stochastic_matrix). 

Re. @jstac's comment, in @spencerlyon2's PR we are still exporting `bellman_operator` so users would still be able to interact with directly for exercises if desired.

A way to remedy this dilemma might be to write up a specification to decide what parts of ``QuantEcon.jl`` need abstraction. For example we might want to class code into pedagogical tools for teaching students (often fairly straight forward, little-to-no abstraction required), versus more involved computational efforts for academics (potentially less straight forward).

My two cents!

jstac:
Hey guys,

I don't have time to give input at the moment.  But given that you're thinking of an implementation that would be compatible with the existing interface, I certainly don't have any issues with you going ahead.

spencerlyon2:
@jstac sounds good. We'll be sure to wait for input before making any _breaking_ changes.

@ZacCranko I'm certainly not sold on the names `solve_vf`, `solve_pf`, `solve_both`. The names of the functions aren't driving the _method_ used to solve the model, rather they distinguish what the return object will be. Meaning, `solve_vf` returns the discretized value function and `solve_pf` returns the discretized policy function (`solve_both` returns value and policy functions).

After sitting on it for the last 6 months, I think we can do better. I still vote we have the `AbstractModel` type, but I think we can do better with our return values. I think I still need to take some time to write out and clarify my ideas, but the basic thought is that we go for just a `solve` that returns a type containing both policy and value functions that we can define other methods on (e.g. perhaps visualization routines). I can see this part going one of three ways (possibly more, I just haven't thought about them yet):

1. The type for all models would be `ModelSolution{TV, TP}` where `TV` is the type of the value function we return and `TP` is the type of the policy function. The `TV` and `TP` would vary depending on the model
2. Define `abstract AbstractSolution` and then have a different `immutable MODELNAMESolution <: AbstractSolution ...` for each model. This gives us ultimate flexibility, but clutters the type system.
3. Define `ModelSolution{TV,TP,TM}`, where `TV` and `TP` are as before, but `TM` is the type of the model. This would just open up more dimensions to dispatch these "other methods" (e.g. plotting) on.

The ideas are still rough, but I think we can explore them to improve the library.

spencerlyon2:
Closing as I don't intend to ever merge this code as is.

We still have the code in a branch, so we aren't losing anything.

Will open a new PR when we have implemented some of the ideas mentioned above. Until then discussion can continue here.

PR #41, opened 2015-04-24
CLN: remove requirement for packages only needed in testing

spencerlyon2:
Creating PR to see what travis does


spencerlyon2:
This didn't work...

Julia internally needs a way to handle dev. vs. user dependencies.

PR #43, opened 2015-05-27
Makes QuantEcon.jl importable

cc7768:
Two fixes:

* Changed `Dict` syntax in mc_tools
* Changed `::(Vector, (Vector, Vector))` to `::(Vector, Tuple{Vector, Vector})`


spencerlyon2:
Travis build failed because the syntax introduced here doesn't work on Julia 0.3.

See https://github.com/JuliaLang/Compat.jl#supported-syntax and update these same lines to work on both 0.3 and 0.4

Issue #16, opened 2014-11-03
Update lqcontrol to handle cross product terms


jstac:
On the Python side lqcontrol.py has been updated to handle cross-product terms

http://quant-econ.net/py/lqcontrol.html#adding-a-cross-product-term

Same functionality required on Julia side.

spencerlyon2:
closed by #48

PR #48, opened 2015-06-19
LQ Cross Product & Revamp

ZacCranko:
This PR adds support for a cross product term in the LQ dynamic programming problem, addressing #16.

Additionally:
* Cleaned up the linear algebra so that algorithms work natively with arrays and scalars without having to convert everything to `Array(T,1,1)` and use broadcast operators everywhere.
* Now `compute_sequence` returns a one dimensional array for scalar problems, and a two dimensional array for vector problems. 

Things to discuss:
* I have renamed the variable for the terminal period (in the finite horizon case) `T` to `term` since `T` is a very special name in the Julia universe, and nearly always used for a type variable. Is this okay?
* I renamed `Rf` to `rf` to cut down on the capital variables floating around, happy to revert this if it offends.

Things not in this PR:
* Should we rename `LQ` to `LinearQuadratic` or `LinearQuadraticControl` in keeping with the julian camel case naming convention for types?
* I have not implemented `Nullable` types for the optional arguments.... I had a quick go at this, but ran out of time. It took me a whole day to get this thing to work and there are other issues that need attending to, but I think the code is a lot cleaner now than before.

N.b. Travis is failing because of some stuff in `robustlq.jl`. If [this line](https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/QuantEcon.jl#L122) is commented out from `QuantEcon.jl` everything behaves itself.

### DIFF
--- a/src/lqcontrol.jl
+++ b/src/lqcontrol.jl
@@ -149,7 +149,7 @@ function _compute_sequence{T}(lq::LQ, x0::Vector{T}, policies)
 	
 	x_path[:,1] = x0
 	u_path[:,1] = -(first(policies)*x0)
-	w_path      =  [dot(lq.C,randn(j)) for i=1:(term+1)]
+	w_path      = [dot(lq.C,randn(j)) for i=1:(term+1)]
 
     for t = 2:term
         f = policies[t]
--- a/src/lqcontrol.jl
+++ b/src/lqcontrol.jl
@@ -20,10 +20,11 @@ type LQ
     R::ScalarOrArray
     A::ScalarOrArray
     B::ScalarOrArray
-    C::Union(Nothing, ScalarOrArray)
+    C::ScalarOrArray
+    N::ScalarOrArray
     bet::Real
     term::Union(Int, Nothing) # terminal period
-    Rf::ScalarOrArray
+    rf::ScalarOrArray
     P::ScalarOrArray
     d::Real
     F::ScalarOrArray # policy rule
@@ -33,54 +34,41 @@ function LQ(Q::ScalarOrArray,
             R::ScalarOrArray,
             A::ScalarOrArray,
             B::ScalarOrArray,
-            C::Union(Nothing, ScalarOrArray)=nothing,
-            bet::ScalarOrArray=1.0,
-            term::Union(Int, Nothing)=nothing,
-            Rf::Union(Nothing, ScalarOrArray)=nothing)
+			C::ScalarOrArray          = zeros(size(R,1)),
+			N::ScalarOrArray          = zero(B'A),
+			bet::ScalarOrArray        = 1.0,
+			term::Union(Int, Nothing) = nothing,
+			rf::ScalarOrArray         = fill(NaN, size(R)...))
+    
     k = size(Q, 1)
     n = size(R, 1)
-
-    if C == nothing
-        j = 1
-        C = zeros(n, j)
-    else
-        j = size(C, 2)
-        if j == 1
-            C = reshape([C], n, j)  # make sure C is a Matrix
-        end
-    end
-
-    if Rf == nothing
-        Rf = fill(NaN, size(R)...)
-    end
-
-    # Reshape arrays to make sure they are Matrix
     F = k==n==1 ? zero(Float64) : zeros(Float64, k, n)
-    P = copy(Rf)
+    P = copy(rf)
     d = 0.0
 
-    LQ(Q, R, A, B, C, bet, term, Rf, P, d, F)
+    LQ(Q, R, A, B, C, N, bet, term, rf, P, d, F)
 end
 
 # make kwarg version
 function LQ(Q::ScalarOrArray,
-            R::ScalarOrArray,
-            A::ScalarOrArray,
-            B::ScalarOrArray;
-            C::Union(Nothing, ScalarOrArray)=nothing,
-            bet::ScalarOrArray=1.0,
-            term::Union(Int, Nothing)=nothing,
-            Rf::Union(Nothing, ScalarOrArray)=nothing)
-    LQ(Q, R, A, B, C, bet, term, Rf)
+			R::ScalarOrArray,
+			A::ScalarOrArray,
+			B::ScalarOrArray,
+			C::ScalarOrArray          = zeros(size(R,1)),
+			N::ScalarOrArray          = zero(B'A);
+			bet::ScalarOrArray        = 1.0,
+			term::Union(Int, Nothing) = nothing,
+			rf::ScalarOrArray         = fill(NaN, size(R)...))
+	LQ(Q, R, A, B, C, N, bet, term, rf)
 end
 
 function update_values!(lq::LQ)
     # Simplify notation
-    Q, R, A, B, C, P, d = lq.Q, lq.R, lq.A, lq.B, lq.C, lq.P, lq.d
+    Q, R, A, B, N, C, P, d = lq.Q, lq.R, lq.A, lq.B, lq.N, lq.C, lq.P, lq.d
 
     # Some useful matrices
     s1 = Q + lq.bet * (B'P*B)
-    s2 = lq.bet * (B'P*A)
+    s2 = lq.bet * (B'P*A) + N
     s3 = lq.bet * (A'P*A)
 
     # Compute F as (Q + B'PB)^{-1} (beta B'PA)
@@ -99,16 +87,16 @@ end
 
 function stationary_values!(lq::LQ)
     # simplify notation
-    Q, R, A, B, C = lq.Q, lq.R, lq.A, lq.B, lq.C
+    Q, R, A, B, N, C = lq.Q, lq.R, lq.A, lq.B, lq.N, lq.C
 
     # solve Riccati equation, obtain P
     A0, B0 = sqrt(lq.bet) * A, sqrt(lq.bet) * B
-    P = solve_discrete_riccati(A0, B0, R, Q)
+    P = solve_discrete_riccati(A0, B0, R, Q, N)
 
     # Compute F
-    S1 = Q + lq.bet * (B' * P * B)
-    S2 = lq.bet * (B' * P * A)
-    F = S1 \ S2
+    s1 = Q + lq.bet * (B' * P * B)
+    s2 = lq.bet * (B' * P * A) + N
+    F = s1 \ s2
 
     # Compute d
     d = lq.bet * trace(P * C * C') / (1 - lq.bet)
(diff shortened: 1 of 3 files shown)


ZacCranko:
N.b. Travis is failing because of some stuff in `robustlq.jl`. If [this line](https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/QuantEcon.jl#L122) is commented out from `QuantEcon.jl` everything behaves itself.

ZacCranko:
Closed as I still want to tweak a couple things

spencerlyon2:
Oh ok.

i was just starting to take a look. I'll wait for the next version.

spencerlyon2:
@ZacCranko is this ready for me to look at?

When I have time (in a few hours) I'll fix the travis issue from robustlq and then we should be able to rely on travis here.

ZacCranko:
@spencerlyon2 Yep this is good to go

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
 
     # Bind states
     lq.P, lq.F, lq.d = P, F, d
-    nothing
 end
 
 function stationary_values(lq::LQ)
-    stationary_values!(lq)
-    return lq.P, lq.F, lq.d
+	_lq = LQ(
```
spencerlyon2: So I see that the python code does update P, F, d when they solve for stationary values. I kinda like that we aren't, but I'm not sure what is correct.

@jstac do you have a suggestion for this?

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
+	x_path = Array(T, n, term+1)
+	u_path = Array(T, n, term)
+	
+	x_path[:,1] = x0
+	u_path[:,1] = -(first(policies)*x0)
+	w_path      = [dot(lq.C,randn(j)) for i=1:(term+1)]
+
+    for t = 2:term
+        f = policies[t]
+        x_path[:,t] = lq.A*x_path[t-1] + lq.B*u_path[t-1] + w_path[t]
```
spencerlyon2: I think there are some bugs here.

We should probably be accessing `(x|u)_path[:, t-1]` instead of just `(x|u)_path[t-1]`.

Looking into fixing it now

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
+
+# dispatch for a vector problem
+function _compute_sequence{T}(lq::LQ, x0::Vector{T}, policies)
+	n, j, term = length(x0), length(lq.C), length(policies)
+	x_path = Array(T, n, term+1)
+	u_path = Array(T, n, term)
+	
+	x_path[:,1] = x0
+	u_path[:,1] = -(first(policies)*x0)
+	w_path      = [dot(lq.C,randn(j)) for i=1:(term+1)]
```
spencerlyon2: This is not correct for cases when n > 1 and j == 1 (see doctoring for python version for explanation of n and j).

Specifically, if n = 2 and j = 1, then we should have `size(C, 1) ==1` and `size(C, 2)== 1`, which means that the variable `w_path` here should be `2 x term+1`. Here it will just be `term+1` b/c `dot` always returns a scalar.

Perhaps `lq.C * randn(j, term+1)` would be better.

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
 
-function compute_sequence(lq::LQ, x0::ScalarOrArray, ts_length=100)
-    # simplify notation
-    Q, R, A, B, C = lq.Q, lq.R, lq.A, lq.B, lq.C
+    x_path, u_path, w_path
+end
+
+# dispatch for a vector problem
+function _compute_sequence{T}(lq::LQ, x0::Vector{T}, policies)
+	n, j, term = length(x0), length(lq.C), length(policies)
```
spencerlyon2: This is not `j`.

`j` should be the number of columns in `C`: `j = size(C, 2)`

spencerlyon2: I'd vote to not name this variable `term` as I got confused in determining whether this was the field or the local at first glance.

Maybe a name like `len`, although I'm certainly open to better suggestions.

src/lqcontrol.jl
spencerlyon2: We will have problems here when, for example `(n, k, j) = (2, 1, 1)`

The problem will be that `lq.B` will probably be a `Vector{Float64}` of length 2 (b/c n=2). Then, `u_path[:, t-1]` will be a `Vector{Float64}` of length k=1.

The correct syntax _for this example_ would be `lq.B .* u_path[:, t-1]`, but that wouldn't work in a `(n, k, j) = (2, 2, 2)` model, for example where we would want the non-broadcasting `*`.

src/lqcontrol.jl
```diff
@@ -16,90 +16,66 @@ http://quant-econ.net/lqcontrol.html
...
+    N::ScalarOrArray
     bet::Real
-    T::Union(Int, Nothing)
-    Rf::Matrix
-    k::Int
-    n::Int
-    j::Int
-    P::Matrix
+    term::Union(Int, Nothing) # terminal period
+    rf::ScalarOrArray
```
spencerlyon2: I see that we moved from `Rf` to `rf`. I know the convention is to have CamelCase for type names, and lower case for variable (field) names.

Sometimes, however, I think it is ok to match the notation of the model (like we do leaving `P` and `F` capitalized).

Any thoughts on this?

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
-    u_path = Array(eltype(x0), lq.k, T)
-    w_path = C * randn(lq.j, T+1)
-
     # Compute and record the sequence of policies
-    policies = Array(typeof(lq.F), T)
-
-    for t=1:T
-        if lq.T != nothing
+    policies = Array(typeof(lq.F), term)
+    for t = 1:term
```
spencerlyon2: I know this was the original structure, but I think we can do better here.

How about we do something like:

```
if lq.term != nothing
    policies = Array(typeof(lq.F), term)
    for t=1:term
        update_values!(lq)
        policies[t] = lq.F
    end
else
    policies = fill(lq.F, term)
end
```

I don't see a need to check  `lq.term != nothing` each iteration.

spencerlyon2:
@ZacCranko, I like the idea of supporting scalar and matrix cases directly. However, going through the code, I found a few errors. 

I see two options:

1. Clean up all the issues I pointed out (and other potential ones) and continue the more Julian way of dispatch and letting the types pick the right implementation
2. Revert back to the old concept of having all fields on our `LQ` type being `Matrix` and do the ugly `reshape` business in the constructor. This approach, although a bit less elegant, is good because then the objects in our field have the dimensions we lay out in the theory -- and we can leverage Julia's overloading of `*` and friends to do the nitty-gritty alignment for us. Oh, and if we go down this route we need to add support for the cross product terms again.

Either way, I think we should make sure al the `examples/*+lq*+.jl` files run properly. We should actually probably create tests cases for each _variant_ of model in those files and make sure the `compute_sequence` function runs on each of them. We would't need to check the answer for accuracy (we already have tests that do that), but we would need to check that models of all `(k, n, j)` being 1 or greater than one work with our implementation.

src/lqcontrol.jl
cc7768: My two cents: I like that `T` is how we usually write things, but understand the problem with doing that in Julia. I'd vote an even smaller deviation by using `capT` although I don't have particularly strong feelings.

spencerlyon2:
One more consideration is that doing it this way makes us effectively write the ` _compute_sequence` algorithm twice. Not a huge deal, but a bit harder to maintain and thought I'd mention it.

src/lqcontrol.jl
ZacCranko: I changed this because I thought since this function is not named with a `!` it shouldn't mutate the input

ZacCranko:
Whoops, I didn't realise we had a heap of examples, I was just checking against the tests. Would probably be a good idea to update the tests with some known results. 

Happy to address the issues you listed above @spencerlyon2.

ZacCranko:
@spencerlyon2 I believe I've sorted out the issues with the `w_path` creation (sorry about that!), though I have not yet checked against the examples. Happy to oblige w.r.t. the naming issues, I'm not sold on `term`, but by concern is that `len` is a bit similar to the function `length`. This of course would not be a catastrophe.

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
+	u_path[:,1] = -(first(policies)*x0)
 
-    for t=1:T
-        if lq.T != nothing
-            update_values!(lq)
-        end
-        policies[t] = lq.F
+    for t = 2:term
+        f = policies[t]
+        x_path[:,t] = lq.A*x_path[t-1] + lq.B*u_path[t-1] + w_path[t]
```
spencerlyon2: @ZacCranko, I thill think there are bugs here. We need to be pulling out `x_path[:, t-1]` and `u_path[:, t-1]`. 

Also, we will still have the issues with dimensionality if we let `lq.B` be a `Vector` and `u_path[:, t-1]` has more than one element.

src/lqcontrol.jl
spencerlyon2: My comment was not so much about the name as it was about the behavior of the function. The python version _does_ mutate the suppled instance of `LQ`, but we aren't here. I'm wondering which approach is better. 

Do we want `P, F, d` to be set to their stationary level, or should they stay at their initial values after this function runs?

src/lqcontrol.jl
```diff
@@ -107,81 +83,103 @@ function update_values!(lq::LQ)
...
+        u_path[:,t] = -(f*x_path[t])
```
spencerlyon2: Same bug here with regards to `x_path[:, t]`

spencerlyon2:
@ZacCranko thanks for working through the `w_path` issues. I think there are still a couple more things we can clean up before this is ready to merge (see new comments).

I liked the suggestion from @cc7768 to use `capT` instead of `term` (or `len`) as the name for the local variable in the `_compute_sequence!` methods.

ZacCranko:
That should do it for mistakes (hopefully). Also `term` is now `capT`

spencerlyon2:
OK, thanks for working through this one. 

I found a couple more bugs, but have taken care of them and things look good. I'm going to rebase this off master and do a force push. So, if you want to hack on this branch again, you need to delete your local copy and pull down a fresh one from here.

spencerlyon2:
Thanks @ZacCranko

ZacCranko:
Excellent, thank you Spencer.

PR #54, opened 2015-06-29
Update mc_convergence_plot.jl

ranjanan:
P needs to be of the type MarkovChain to go into mc_compute stationary. You mentioned you thought about adding the old method back right? But the MarkovChain constructor would be useful in checking whether the array satisfies the condition sum(P[i,:]) = 1. Or do you want to integrate that into every MarkovChain routine?

### DIFF
--- a/examples/mc_convergence_plot.jl
+++ b/examples/mc_convergence_plot.jl
@@ -30,5 +30,6 @@ end
 
 ax[:scatter](x_vals, y_vals, z_vals, c="r", s = 60)
 
+P = MarkovChain(P)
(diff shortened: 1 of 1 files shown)


spencerlyon2:
@ranjanan thanks for pointing this out -- we should definitely fix it.

I also like the comment you made here about using the `MarkovChain` as a way to enforce the invariants imposed by the inner constructor. That's probably a good approach to take for all the functions defined in `mc_tools` -- but making that change would require more work to be done on the website. I'll log an issue there.

PR #56, opened 2015-07-08
New features for mc_tools

ZacCranko:
* Moved to LightGraphs for graph calculations
* New functions added

`recurrent_classes(mc::MarkovChain)`
`communication_classes(mc::MarkovChain)`
`is_irreducible(mc::MarkovChain)`
`period(mc::MarkovChain)`

Tests for these new functions are not included in this pull. I'm going to change up the tests file, so I'm submitting this pull and making those modifications separately, but all the old tests pass fine with the new code.

### DIFF
--- a/src/mc_tools.jl
+++ b/src/mc_tools.jl
@@ -10,10 +10,8 @@ References
 
 http://quant-econ.net/jl/finite_markov.html
 =#
-
-# new method to check if all elements of an array x are equal to p
-isapprox(x::Array, p::Number) = all([isapprox(x[i], p) for i=1:length(x)])
-isapprox(p::Number, x::Array) = isapprox(x, p)
+import LightGraphs: DiGraph, period, attracting_components,
+                    strongly_connected_components, is_strongly_connected
 
 """
 Finite-state discrete-time Markov chain.
@@ -30,15 +28,19 @@ positive, and all rows must sum to unity
 type MarkovChain{T<:Real}
     p::Matrix{T} # valid stochastic matrix
 
-    function MarkovChain(p)
+
+    function MarkovChain(p::Matrix)
         n, m = size(p)
 
-        n != m && throw(ArgumentError("stochastic matrix must be square"))
-        any(p .< 0) &&
+        if n != m
+            throw(ArgumentError("stochastic matrix must be square"))
+        elseif any(p.<0)
             throw(ArgumentError("stochastic matrix must have nonnegative elements"))
-        isapprox(sum(p, 2), one(T)) ||
+        elseif any(x->!isapprox(sum(x), 1), [p[i, :] for i = 1:n])
             throw(ArgumentError("stochastic matrix rows must sum to 1"))
-        new(p)
+        end
+
+        return new(p)
     end
 end
 
@@ -54,6 +56,7 @@ function Base.show(io::IO, mc::MarkovChain)
     println(io, mc.p)
 end
 
+# solve x(P-I)=0 by eigen decomposition
 function eigen_solve{T}(p::Matrix{T})
     ef = eigfact(p')
     isunit = map(x->isapprox(x, 1), ef.values)
@@ -66,7 +69,7 @@ function eigen_solve{T}(p::Matrix{T})
     x
 end
 
-# function to solve x(P-I)=0 by lu decomposition
+# solve x(P-I)=0 by lu decomposition
 function lu_solve{T}(p::Matrix{T})
     n, m = size(p)
     x   = vcat(Array(T, n-1), one(T))
@@ -82,37 +85,34 @@ function lu_solve{T}(p::Matrix{T})
     x
 end
 
-gth_solve{T<:Integer}(A::Matrix{T}) = gth_solve(convert(Array{Float64, 2}, A))
+gth_solve{T<:Integer}(a::Matrix{T}) = gth_solve(convert(Array{Rational, 2}, a))
 
-function gth_solve{T<:Real}(A::AbstractMatrix{T})
-    A1 = copy(A)
-    n = size(A1, 1)
+# solve x(P-I)=0 by the GTH method
+function gth_solve{T<:Real}(original::Matrix{T})
+    a = copy(original)
+    n = size(a, 1)
     x = zeros(T, n)
 
-    # === Reduction === #
-    for k in 1:n-1
-        scale = sum(A1[k, k+1:n])
-        if scale <= 0
-            # There is one (and only one) recurrent class contained in
-            # {1, ..., k};
-            # compute the solution associated with that recurrent class.
+    @inbounds for k in 1:n-1
+        scale = sum(a[k, k+1:n])
+        if scale <= zero(T)
             n = k
             break
         end
-        A1[k+1:n, k] ./= scale
+        a[k+1:n, k] /= scale
 
         for j in k+1:n, i in k+1:n
-            A1[i, j] += A1[i, k] * A1[k, j]
+            a[i, j] += a[i, k] * a[k, j]
         end
     end
 
-    # === Backward substitution === #
+    # backsubstitution
     x[end] = 1
-    for k in n-1:-1:1, i in k+1:n
-        x[k] += x[i] * A1[i, k]
+    @inbounds for k in n-1:-1:1, i in k+1:n
+        x[k] += x[i] * a[i, k]
     end
 
-    # === Normalization === #
+    # normalisation
     x / sum(x)
 end
 
@@ -133,7 +133,7 @@ algorithm presented by Grassmann-Taksar-Heyman (GTH)
 The following references were consulted for the GTH algorithm
 
 - W. K. Grassmann, M. I. Taksar and D. P. Heyman, "Regenerative Analysis and
-Steady State Distributions for Markov Chains," Operations Research (1985),
+Steady State Distributions for Markov Chains, " Operations Research (1985),
 1107-1116.
 - W. J. Stewart, Probability, Markov Chains, Queues, and Simulation, Princeton
 University Press, 2009.
@@ -142,7 +142,7 @@ University Press, 2009.
 [eigen_solve, lu_solve, gth_solve]
 
 """
-Find the irreducible subsets of the `MarkovChain`
+Find the recurrent classes of the `MarkovChain`
 
 ##### Arguments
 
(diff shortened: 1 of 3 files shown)


oyamad:
@ZacCranko Thanks, it looks very nice.

0. It will be helpful to have [`is_aperiodic`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/mc_tools.py#L208) as well:

  ```
  is_aperiodic(mc::MarkovChain) = period(DiGraph(mc.p)) == 1
  ```

  or add [`is_aperiodic`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/graph_tools.py#L255) to `LightGraphs.jl` and do

  ```
  is_aperiodic(mc::MarkovChain) = is_aperiodic(DiGraph(mc.p))
  ```

2. In [`mc_tools.py`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/mc_tools.py#L215), the period is also defined for a reducible Markov chain, to be the LCM of the periods of the recurrent classes. I think it is better to adopt the same definition between the two versions.

3. Also remaining is [`cyclic_classes`](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/mc_tools.py#L230).

4. For `method` in [`mc_compute_stationary`](https://github.com/QuantEcon/QuantEcon.jl/blob/cf59a18c90c56f9a55a4157771306a4ebb8f9a76/src/mc_tools.jl#L178), by what criterion should the user choose among `gth`, `lu`, and `eigen` (what is the point of having three methods)?

src/mc_tools.jl
```diff
@@ -29,24 +26,24 @@ positive, and all rows must sum to unity
...
+        throw(ArgumentError("stochastic matrix must have nonnegative elements"))
+    elseif any(x->!isapprox(sum(x),1), [p[i,:] for i in 1:n])
+        throw(ArgumentError("stochastic matrix rows must sum to 1"))
     end
-end
 
-# Provide constructor that infers T from eltype of matrix
-MarkovChain{T<:Real}(p::Matrix{T}) = MarkovChain{T}(p)
+    return MarkovChain{T}(p)
+end
```
spencerlyon2: We should restore the inner+outer constructor pair. The reason is that we never want to be able to create an instance of `MarkovChain` without these 3 conditions holding -- these are the invariants referred to in the [inner constructor](http://julia.readthedocs.org/en/latest/manual/constructors/#inner-constructor-methods) section of the mainual

src/mc_tools.jl
```diff
@@ -54,65 +51,63 @@ function Base.show(io::IO, mc::MarkovChain)
...
 end
 
+# solve x(P-I)=0 by eigen decomposition
 function eigen_solve{T}(p::Matrix{T})
     ef = eigfact(p')
-    isunit = map(x->isapprox(x, 1), ef.values)
+    isunit = map(x->isapprox(x,1), ef.values)
     x = real(ef.vectors[:, isunit])
-    x ./= sum(x, 1) # normalisation
+    x ./= sum(x,1) # normalisation
```
spencerlyon2: To follow the style of the rest of the library, a comma must _always_ be followed by a space.

This applies multiple times in this PR, but I will only mention it here to preserve everyone's inbox!

oyamad:
WIth Version 0.3.10,  I got the following error:

```julia
julia> using QuantEcon

julia> mc = MarkovChain([0.4 0.6; 0.2 0.8])
Discrete Markov Chain
stochastic matrix:
2x2 Array{Float64,2}:
 0.4  0.6
 0.2  0.8


julia> mc_compute_stationary(mc)
ERROR: error compiling mc_compute_stationary: error compiling __mc_compute_stationary#140__: unsupported or misplaced expression => in function __mc_compute_stationary#140__
```

It works fine with Version 0.4.0.

spencerlyon2:
Thanks for the note. We'll take care of it

src/mc_tools.jl
```diff
@@ -197,30 +175,24 @@ of the three `_solve` methods to use.
 distribution of `mc.p`
 
 """
-function mc_compute_stationary{T}(mc::MarkovChain{T}; method::Symbol=:gth)
-    @compat solvers = Dict(:gth => gth_solve,
-                           :lu => lu_solve,
-                           :eigen => eigen_solve)
+function mc_compute_stationary{T}(mc::MarkovChain{T}; method=:gth)
+    solvers = Dict(:gth => gth_solve, :lu => lu_solve, :eigen => eigen_solve)
```
spencerlyon2: This line needs `@compat` to work on 0.3 -- as @oyamad noted

spencerlyon2: We should also make sure that the `method` argument is typed as `::Symbol`. 

(we should restore exactly what was here originally -- it reverts both changes)

src/mc_tools.jl
```diff
@@ -10,10 +10,7 @@ References
 
 http://quant-econ.net/jl/finite_markov.html
 =#
-
-# new method to check if all elements of an array x are equal to p
-isapprox(x::Array, p::Number) = all([isapprox(x[i], p) for i=1:length(x)])
-isapprox(p::Number, x::Array) = isapprox(x, p)
+import LightGraphs: DiGraph, period, attracting_components, strongly_connected_components
```
spencerlyon2: We need to make a decision regarding imports.

For the rest of the code we have put all `using` statements in the main module file (`src/QuantEcon.jl`). 

Here it is a little different because you are adding a new method to the `period` function. 

I think I would prefer that we do `using LightGraphs: DiGraph, attracting_components, strongly_connected_components` in (`src/QuantEcon.jl`) and then define the method below 

```julia
function LightGraphs.period(mc::MarkovChain)
    ...
```

I'm not 100% convinced of this, what thoughts do others have?

If we do make the proposed decision, I will probably not `import` anything from `Base` and add new methods using the syntax `Base.function_name` -- over time I find that I am preferring this explicit form of adding new methods

src/mc_tools.jl
```diff
@@ -289,10 +261,16 @@ None modifies `samples` in place
...
         samples[t] = rand(dist[samples[t-1]])
     end
     nothing
 end
+
+function simulate(mc::MarkovChain, 
+    init::Vector = vcat(1,zeros(n_states(mc)-1)), 
+    sample_size=1000)
+    return mc_sample_path(mc,init,sample_size)
+end
```
spencerlyon2: Need a new line at the end of the file

spencerlyon2: Need a new line at the end of the file

spencerlyon2: Another picky style thing -- have all function arguments that are on different lines start at the same column as the first function argument like we do [here](https://github.com/QuantEcon/QuantEcon.jl/blob/cf59a18c90c56f9a55a4157771306a4ebb8f9a76/src/mc_tools.jl#L240-L242) (and all other places in the library)

Sorry about these irritating requests. I really do need to write up a style guide so people besides me know what the style guidelines are. I'll open up an issue to remind myself.

ZacCranko: Happy to follow a style guide.

src/mc_tools.jl
ZacCranko: My philosophy would be to locate import statements that multiple sections of the library use in `QuantEcon.jl` and locate import statements that only a couple packages use in their own files. 
Since we're not providing any functionality to the user such that they interact with any of the other types or methods in LightGraphs, this is why i chose `import` over `using`.

ZacCranko:
Thanks guys, I believe I've addressed everything except for the `import` question,  and `cyclic_components` which I'm not gonna tackle just yet

oyamad:
@ZacCranko Thanks!

* `mc_compute_stationary` returns a 1-dimensional array if `mc` has a unique recurrent class (in which case there is a unique stationary state) and a 2-dimensional array otherwise, while it is always 1-dimensional in the Python version. Would there be any problem having different behaviors?

* What about the methods for `mc_compute_stationary`? What is your intention?

* This is not about this PR itself, but I think `mc_sample_path(mc, init, sample_size)` should return an array of length `sample_size`, not `sample_size+1`.

ZacCranko:
I don't really have much of an opinion on the solution methods. Though since this is a pedagogical tool, there is probably utility in allowing the the user to compare GTH and the other methods, especially if we, in the future, were to have a lecture on approaches to compute the stationary distribution of a Markov chain. This would be consistent with the fact that we're exporting `gth_solve`.

oyamad:
I see, thanks. Makes sense for education purposes.

ZacCranko:
Thanks for that, good to know! Comments fixed.

oyamad:
Another issue: Why does `mc_compute_stationary` return stationary distributions as *columns*, whereas `MarkovChain` accepts a *row*-stochastic matrix, for which a stationary distribution is a *left*-eigenvector? I find this somewhat inconsistent.

ZacCranko:
I'm happy to have `mc_compute_stationary` return what ever anyone would like. However since Julia is column-major it would be more efficient computationally (for accessing) if vectors are columns.

oyamad:
Right, then we might consider having `MarkovChain` accept a column-stochastic matrix. Mathematically, there is no reason for one of row-stochastic or column-stochastic to be better than the other.

ZacCranko:
The only convenience with the current orientation of the stochastic matrix is that it coincides with the definition of an adjacency matrix.

oyamad:
Ah that's a good point.

Would it be costly to transpose the column-stochastic input and pass it to `DiGraph`?

ZacCranko:
Computing the transpose of a matrix is O(n), but this is somewhat unrelated to what this PR is addressing. If you really feel strongly about this I think it would be best to open an issue.

oyamad:
Yes, that's right.

jstac:
Sorry if I get something wrong here as I'm just going by the thread, but I would recommend that we return the stationary distributions -- and all other distributions -- as row vectors, either by transposing them as they are passed out or some other approach.  It's entirely conventional in the field of Markov chains that distributions are row vectors.  Similarly, the matrix should be such that P[i, j] means prob of moving from state i to state j.  (In fact that's why distributions end up being row vectors that we multiply from the left.)  So the challenge is to implement this standard interface in the most efficient way.

spencerlyon2:
What if we return the stationary distributions as a `Vector{Vector{Float64}`? Then we don't have to worry about orientation

jstac:
I guess a flat array is fine.

ZacCranko:
Ahoy hoy, is this ready to be merged?

spencerlyon2:
Hey sorry I kinda forgot we hadn't merged this. 

I would love to see some tests of the new methods. i realize you want to change the tests. Will that happen soon?

ZacCranko:
Yep, that'll happen pretty quick.

spencerlyon2:
Great, I'll break my rule of merging without tests this time...

Thanks for the work on this.

PR #60, opened 2015-07-21
fixed mc solution notebook

jstac:
Fixed the finite MC solution notebook to the new style of creating MarkovChain instances and calling MC related functions.  Was broken but now runs.


spencerlyon2:
Thanks @jstac!

PR #71, opened 2015-09-01
change unicode to ascii for latex parsing

mmcky:
@spencerlyon2 This is a small change to allow the lectures to compile the ``pdf`` file. The unicode character was causing compilation issues.


spencerlyon2:
Thanks!

PR #72, opened 2015-09-04
Arellano

cc7768:
This brings code in to do the Arellano value function iteration. It is almost identical to the Python versions of the code.

We might want to hold off on merging until I've (or someone else has) had a chance to write/clean the docs. Just opening as a place holder.

### DIFF
--- a/src/models/arellano_vfi.jl
+++ b/src/models/arellano_vfi.jl
@@ -30,7 +30,7 @@ ny : int
 nB : int
     Number of points in B grid
 =#
-immutable Arellano_Economy
+immutable ArellanoEconomy
     # Model Parameters
     β::Float64
     γ::Float64
@@ -56,7 +56,7 @@ immutable Arellano_Economy
     defprob::Array{Float64, 2}
 end
 
-function Arellano_Economy(;β=.953, γ=2., r=0.017, ρ=0.945, η=0.025, θ=0.282,
+function ArellanoEconomy(;β=.953, γ=2., r=0.017, ρ=0.945, η=0.025, θ=0.282,
                           ny=21, nB=251)
 
     # Create grids
@@ -74,14 +74,14 @@ function Arellano_Economy(;β=.953, γ=2., r=0.017, ρ=0.945, η=0.025, θ=0.282
     q = ones(nB, ny) .* (1 / (1 + r))
     defprob = Array(Float64, nB, ny)
 
-    return Arellano_Economy(β, γ, r, ρ, η, θ, ny, nB, ygrid, ydefgrid, Bgrid, Π,
+    return ArellanoEconomy(β, γ, r, ρ, η, θ, ny, nB, ygrid, ydefgrid, Bgrid, Π,
                             vf, vd, vc, policy, q, defprob)
 end
 
-u(ae::Arellano_Economy, c) = c^(1 - ae.γ) / (1 - ae.γ)
-_unpack(ae::Arellano_Economy) =
+u(ae::ArellanoEconomy, c) = c^(1 - ae.γ) / (1 - ae.γ)
+_unpack(ae::ArellanoEconomy) =
     ae.β, ae.γ, ae.r, ae.ρ, ae.η, ae.θ, ae.ny, ae.nB
-_unpackgrids(ae::Arellano_Economy) =
+_unpackgrids(ae::ArellanoEconomy) =
     ae.ygrid, ae.ydefgrid, ae.Bgrid, ae.Π, ae.vf, ae.vd, ae.vc, ae.policy, ae.q, ae.defprob
 
 
@@ -95,7 +95,7 @@ at every state by solving for the optimal choice of savings
 
 Note: Updates value functions in place.
 =#
-function one_step_update!(ae::Arellano_Economy, EV, EVd, EVc)
+function one_step_update!(ae::ArellanoEconomy, EV, EVd, EVc)
 
     # Unpack stuff
     β, γ, r, ρ, η, θ, ny, nB = _unpack(ae)
@@ -136,7 +136,7 @@ function one_step_update!(ae::Arellano_Economy, EV, EVd, EVc)
     nothing
 end
 
-function compute_prices!(ae::Arellano_Economy)
+function compute_prices!(ae::ArellanoEconomy)
     # Unpack parameters
     β, γ, r, ρ, η, θ, ny, nB = _unpack(ae)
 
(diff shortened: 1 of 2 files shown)


jstac:
@cc7768 Great, thanks.  Do you know why the build is failing?

spencerlyon2:
It looks like all the travis errors are associated with many of the model solutions not being _roughly_ (tolerance `1e-12`) fixed points. I'm not really sure why this would have changed. Maybe it is time to just regenerate stale cached solution files? Tests would take a bit longer to run in that case, but we would be testing all the routines as they currently are.

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,261 @@
...
+    copy!(ae.q, (1 - ae.defprob) / (1 + r))
+
+    nothing
+end
+
+#=
+This performs value function iteration and stores all of the data inside
+the ArellanoEconomy type.
+=#
+function vfi!(ae::Arellano_Economy; tol=1e-8, maxit=10000)
```
spencerlyon2: Could we re-use the vfi concept from all other models:

- Provide `bellman_operator`
- Call `compute_fixed_point` on that operator. 

I haven't read what you are doing here closely enough to see if it is vastly different from what we do for other models.

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,261 @@
...
+eta : float
+    Standard deviation of the income process
+theta : float
+    Probability of re-entering financial markets in each period
+ny : int
+    Number of points in y grid
+nB : int
+    Number of points in B grid
+=#
+immutable Arellano_Economy
```
spencerlyon2: No underscore in the type name

spencerlyon2:
Another general comment:

I see here that we are including the vfi and solution grids as part of the type. We don't really do this for any of the other models. Can you explain why we do this (i.e. what gains we have from setting it up this way instead of how we think about the other models)? I'm not sure if it got across, but my tone here is more inquisitive than acusitory -- I'm happy to use the current structure if we believe it simplifies things.

jstac:
I think that's [pretty common](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/models/lucastree.py) in what we've done so far, at least on the Python side.  Typically a good grid depends on the parameters --- for example, large variance in shocks usually requires a larger grid.  So it seems natural to store the grid together with the parameters.

I'm happy to hear other sides of the argument.

I sometimes think we should write up the elements of `quantecon/models` as scripts without fancy types, and make all the variables global.  This is attractive for some users -- people who will look at our code and inevitably say "I could do that with half as many lines in MATLAB".  But somehow I always end up writing classes.

cc7768:
Thanks for all the comments!

* We can probably use the `compute_fixed_point!` and `bellman_operator!` approach. It shouldn't be too difficult to incorporate that approach and it is more consistent with our own features.
* Providing all of the value functions and policy functions inside the type was geared towards making all the functions have relatively short lists of arguments (which in my opinion makes them easier to interact with). Calling `simulate` is then as simple as `simulate(ae, 1000)`. Another option would be to have two types, but for the sake of simplicity it might be nice to stick with one and just leave the solution concepts inside it.
* John, the scripts would be nice to show that we could do it in less lines via script, but I think it would be better if we could use the OOP section to help people understand why this is a strict improvement. In particular, it would be great to communicate that using types (and generally breaking up algorithms into pieces with functions) around gives us the power to write shorter algorithms (that look like how the algorithms are presented int he paper) even if it takes requires us to do some pre-processing and organization.

src/models/arellano_vfi.jl
cc7768: Fixed this one

cc7768:
I gave myself an allotted 20 minutes to get a start on the docstrings before I moved on to some other work. Pushed what I have -- They are mostly done.

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,333 @@
...
+    q::Array{Float64, 2}
+    defprob::Array{Float64, 2}
+end
+
+"""
+This is the default constructor for building an economy as presented
+in Arellano 2008.
+
+##### Arguments
+* `β::Real` : Time discounting parameter
```
spencerlyon2: Please revise to include the `;` denoting keyword argument and put the default in parens. So this line should read:

```
`;β::Real(0.953)` : Time discounting parameter
```

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,333 @@
...
+This function performs the one step update of the value function for the
+Arellano model-- Using current value functions and their expected value,
+it updates the value function at every state by solving for the optimal
+choice of savings
+
+##### Arguments
+
+* `ae::ArellanoEconomy` : This is the economy we would like to update the
+                          value functions for
+* `EV::Array{Float64, 2}` : Expected value function at each state
```
spencerlyon2: If these types are required, let's add them to the actual method definition

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,333 @@
...
+    nothing
+end
+
+"""
+This performs value function iteration and stores all of the data inside
+the ArellanoEconomy type.
+
+##### Arguments
+
+* `ae::ArellanoEconomy` : This is the economy we would like to solve
```
spencerlyon2: Same note about `;` and `(default)`

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,333 @@
...
+end
+
+"""
+This function simulates the Arellano economy
+
+##### Arguments
+
+* `ae::ArellanoEconomy` : This is the economy we would like to solve
+* `T::Int` : Number of periods to simulate
+* `y_init::Float64` : The level of income we would like to start with
```
spencerlyon2: Again the `;` and `(default)` (do I sound like a broken record :) )

src/models/arellano_vfi.jl
```diff
@@ -0,0 +1,333 @@
...
+* `y_sim_val::Vector{Float64}` : Simulated values of income
+* `q_sim_val::Vector{Float64}` : Simulated values of prices
+* `default_status::Vector{Float64}` : Simulated default status (True if in default)
+
+##### Notes
+
+* This updates all value functions, policy functions, and prices in place.
+
+"""
+function QuantEcon.simulate(ae::ArellanoEconomy, T::Int=5000;
```
spencerlyon2: Let's rename `T` to `capT`. As @ZacCranko once pointed out, there is a very strong convention to have `T` be a type parameter in Julia code

spencerlyon2:
Do we have tests for this code?

mmcky:
See #73.

PR #73, opened 2015-09-04
ENH: added uncertainty traps code

spencerlyon2:
This isn't quite ready to merge.

It totally replicates the python, but I would like to add docstrings and tests before merging. Any help with those two items would be appreciated.


jstac:
@spencerlyon2 Thanks!

mmcky:
@spencerlyon2 Having a look at the file there appears to be some documentation explaining the variables. But it is not in PEP8 etc. Is this what you are looking for - transfer the inline comments to class doctstrings with parameters etc?

spencerlyon2:
Given the impending changes to the model structure, I'm not so worried about tests/docs anymore.

I think this can be merged as is

mmcky:
@spencerlyon2 Ok I will keep an eye out for the merge - then I will update the two lectures on ``arellano`` and ``uncertainty_traps`` to make sure the ``julia`` ones are working. We can then do an update to the website and open a new project for relocating the ``models`` sub-package

PR #76, opened 2015-11-03
random_probvec: Remove temporary array

oyamad:
Corresponding to QuantEcon/QuantEcon.py#184


spencerlyon2:
Looks good to me . If Travis builds on 0.4 we can merge

Issue #52, opened 2015-06-28
Error message in mc_sample_path


ranjanan:
I receive an error message on using mc_sample_path()

julia> using QuantEcon

julia> P = [0.4 0.6; 0.8 0.2]
2x2 Array{Float64,2}:
 0.4  0.6
 0.8  0.2

julia> P = MarkovChain(P)
Discrete Markov Chain
stochastic matrix:
2x2 Array{Float64,2}:
 0.4  0.6
 0.8  0.2

julia> mc_sample_path(P, [0.5, 0.5], 100000)
ERROR: type cannot be constructed
 in map at abstractarray.jl:1328

I'm using Julia version 0.3.9 on Mac OS X 10.10.3

ZacCranko:
Working fine here on v0.3.10, can you try updating Julia then running `Pkg.update()` and see if this still persists?

oyamad:
Reproduced the same error message with version 0.3.10 + QuantEcon 0.1.2 (on Mac OS X 10.9.5),
while it works fine with 0.4.0-dev.

ZacCranko:
@oyamad can you run `Pkg.update()` on your v0.3 since we're calling some things from `Distributions.jl` that I want to exclude as potential causes.

spencerlyon2:
Thanks for reporting. I'm considering re-writing this routine anyway as I think we can make it more performant. Implementing an algorithm like what we just did in the numba version (the one with embedded bisection) may be faster than calling out to Categorical from Distributions.jl

oyamad:
@ZacCranko Got the same error.

ZacCranko:
I'm unable to reproduce this on v0.3.10 or v0.4.0-dev+5645

spencerlyon2:
Scratch my previous comment. I was confusing the new implementation of this algorithm with a different implementation I saw earlier.

I still wouldn't be surprised if using `searchsortedfirst` was more efficient than calling Categorical. Maybe we just write up an implementation and run some benchmarks against the two?

ranjanan:
@ZacCranko Just as @oyamad said, I'm getting the same error on Julia v0.3.10 on Mac OS X 10.10.3 as well. On a 16 day old master of Julia , I got this error while trying "using QuantEcon"

 using QuantEcon
Warning: New definition 
    serialize(Any, DataStructures.HashDict) at /Users/ranjan/.julia/v0.4/DataStructures/src/hashdict.jl:91
is ambiguous with: 
[...]

PR #77, opened 2015-11-05
MarkovChain simulate: Match with python version

oyamad:
1.
This is to propose matching the `simulate` method with that in the Python version (see https://github.com/QuantEcon/QuantEcon.py/issues/146#issuecomment-135328616). It works as follows:

```julia
julia> using QuantEcon

julia> mc = MarkovChain([0.4 0.6; 0.2 0.8])
Discrete Markov Chain
stochastic matrix:
2x2 Array{Float64,2}:
 0.4  0.6
 0.2  0.8


julia> simulate(mc, 5)
5-element Array{Int64,1}:
 2
 2
 1
 1
 2

julia> simulate(mc, 5, [1, 2])
5x2 Array{Int64,2}:
 1  2
 1  1
 2  1
 2  1
 2  2

julia> simulate(mc, 5, num_reps=3)
5x3 Array{Int64,2}:
 1  2  2
 2  1  1
 2  2  1
 1  2  2
 1  1  1

julia> simulate(mc, 5, 1, num_reps=3)
5x3 Array{Int64,2}:
 1  1  1
 1  1  2
 2  2  2
 2  2  2
 1  2  2

julia> simulate(mc, 5, [1, 2], num_reps=3)
5x6 Array{Int64,2}:
 1  2  1  2  1  2
 2  2  1  1  2  1
 2  1  2  2  1  2
 2  2  1  2  1  2
 2  2  2  2  2  2

```

2.
As discussed in https://github.com/QuantEcon/QuantEcon.jl/issues/52#issuecomment-116389128, it uses `searchsortedfirst` (via [`DiscreteRV`](https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/discrete_rv.jl)). The method using `searchsortedfirst` (`simulate`) seems faster than the previous one using `Categorical` (`mc_sample_path`) for medium- and large-size Markov chains; see [mc_simulate.jl.ipynb](https://nbviewer.jupyter.org/gist/oyamad/34477928fdc330c9da5b/mc_simulate.jl.ipynb).

In passing, Numba seems to do a good job; see [mc_simulate.py.ipynb](https://nbviewer.jupyter.org/gist/oyamad/34477928fdc330c9da5b/mc_simulate.py.ipynb).

3.
I am not sure if I properly exploited the multiple dispatch system of Julia. Any suggestion/correction etc will be welcome!

### DIFF
--- a/src/mc_tools.jl
+++ b/src/mc_tools.jl
@@ -306,19 +306,24 @@ The sample path from the `j`-th repetition of the simulation with initial state
 - `mc::MarkovChain` : MarkovChain instance.
 - `ts_length::Int` : Length of each simulation.
 - `init::Vector{Int}` : Vector containing initial states.
-- `;num_reps::Int(1)` : Number of repetitions of simulation.
+- `;num_reps::Union{Int, Void}(nothing)` : Number of repetitions of simulation.
 
 ##### Returns
 
 - `X::Array{Int}` : Array containing the sample paths as columns, of shape
-(ts_length, k), where k = length(init)*num_reps.
+(ts_length, k), where k = length(init) if num_reps=nothing, and k =
+length(init)*num_reps otherwise.
 
 """
 function simulate(mc::MarkovChain,
                   ts_length::Int,
                   init::Vector{Int};
-                  num_reps::Int=1)
-    k = length(init) * num_reps
+                  num_reps::Union{Int, Void}=nothing)
+    k = length(init)
+    if num_reps == nothing
+        num_reps = 1
+    end
+    k *= num_reps
     X = Array(Int, ts_length, k)
     X[1, :] = repmat(init, num_reps)
 
--- a/src/mc_tools.jl
+++ b/src/mc_tools.jl
@@ -303,24 +303,19 @@ Simulate time series of state transitions of the Markov chain `mc`.
 - `mc::MarkovChain` : MarkovChain instance.
 - `ts_length::Int` : Length of each simulation.
 - `init::Vector{Int}` : Vector containing initial states.
-- `;num_reps::Union{Int, Void}(nothing)` : Number of repetitions of simulation.
+- `;num_reps::Int(1)` : Number of repetitions of simulation.
 
 ##### Returns
 
 - `X::Array{Int}` : Array containing the sample paths as columns, of shape
-(ts_length, k), where k = length(init) if num_reps=nothing, and k =
-length(init)*num_reps otherwise.
+(ts_length, k), where k = length(init)*num_reps.
 
 """
 function simulate(mc::MarkovChain,
                   ts_length::Int,
                   init::Vector{Int};
-                  num_reps::Union{Int, Void}=nothing)
-    k = length(init)
-    if num_reps == nothing
-        num_reps = 1
-    end
-    k *= num_reps
+                  num_reps::Int=1)
+    k = length(init) * num_reps
     X = Array(Int, ts_length, k)
     X[1, :] = repmat(init, num_reps)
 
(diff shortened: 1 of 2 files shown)


src/mc_tools.jl
```diff
@@ -295,8 +295,115 @@ function mc_sample_path!(mc::MarkovChain, samples::Array)
...
+
+- `X::Array{Int}` : Array containing the sample paths as columns, of shape
+(ts_length, k), where k = length(init) if num_reps=nothing, and k =
+length(init)*num_reps otherwise.
+
+"""
+function simulate(mc::MarkovChain,
+                  ts_length::Int,
+                  init::Vector{Int};
+                  num_reps::Union{Int, Void}=nothing)
```
spencerlyon2: No need for the union here and nothing here. Just set a default argument of `num_reps=1`

src/mc_tools.jl
```diff
@@ -295,8 +295,115 @@ function mc_sample_path!(mc::MarkovChain, samples::Array)
...
+##### Returns
+
+- `X::Array{Int}` : Array containing the sample path(s) as column(s), of shape
+(ts_length,) if num_reps=nothing; of shape (ts_length, num_reps) otherwise.
+
+"""
+function simulate(mc::MarkovChain,
+                  ts_length::Int,
+                  init::Int;
+                  num_reps::Union{Int, Void}=nothing)
```
spencerlyon2: Same comment, just use `num_reps::Int=1`

src/mc_tools.jl
```diff
@@ -295,8 +295,115 @@ function mc_sample_path!(mc::MarkovChain, samples::Array)
...
+(ts_length,) if num_reps=nothing; of shape (ts_length, num_reps) otherwise,
+where the initial state is randomly drawn for each simulation.
+
+"""
 function simulate(mc::MarkovChain,
-                  init::Vector = vcat(1, zeros(n_states(mc)-1)),
-                  sample_size=1000)
-    return mc_sample_path(mc, init, sample_size)
+                  ts_length::Int;
+                  num_reps::Union{Int, Void}=nothing)
```
spencerlyon2: again...

spencerlyon2:
Looks great overall, thanks for adding this.

I have another request: can we document how the columns are ordered when `num_reps > 1`?

src/mc_tools.jl
oyamad: @spencerlyon2 Thank you for the comments.

Here I wanted to follow the convention in the python version, where the returned array is 1-dimensional when `init` is a scalar and `num_reps` is not specified, while it is 2-dimensional when `num_reps` is specified (even if `num_reps = 1`).

oyamad:
"If `init` is a vector, the sample path from the `j`-th repetition of the simulation with initial state `init[i]` is stored in the `i*j`-th column of the matrix `X`."
 
Is this what you mean?

spencerlyon2:
Sorry this has sat for so long. 

@oyamad what's the status on this, do you recall?

src/mc_tools.jl
oyamad: Done.

oyamad: Reverted.

oyamad:
@spencerlyon2 Sorry I had forgotten about this PR.

I now remembered why I did `num_reps::Union{Int, Void}=nothing` in the first `simulate` method; it is called from the other methods with `num_reps=nothing`. There may be a good way to avoid it, but I left it as is. Anyway, the issue is whether it is a good implementation that if `num_reps` is not supplied, the output is a 1-dimensitonal array, while if it is supplied, the output is 2-dimensional even when `num_reps=1`.

spencerlyon2:
Ahh that's right thanks @oymad.

The behavior you described will be type stable, so I think it is a good choice.  I'm happy to merge as is. 

Thanks

oyamad:
@spencerlyon2 Thanks for taking care of the tests.

spencerlyon2:
You're welcome, thanks again for the quality work!

PR #78, opened 2015-11-05
Add Discrete DP

mmcky:
**Status:** Not Ready to be Merged, Open for Review

Add Discrete DP
  - value iteration
  - policy function iteration 
  - modified policy function iteration

Outstanding items
  - [x] Preliminary tests have been translated but do not yet run. They need a ``julian`` review. The file is currently documented with ``TODO``.
  - [ ] State-Action Pair Formulation
  - [ ] Sparse Matrix Support
  - [x] Docstring cleanup

I will submit an associated PR for the QuantEcon website to add ``julia`` lecture

### DIFF
--- a/src/markov/ddp.jl
+++ b/src/markov/ddp.jl
@@ -1,0 +1,468 @@
+#=
+Discrete Decision Processes
+
+@author : Daisuke Oyama (Ported by Spencer Lyon, Matthew McKay)
+
+@date: 24/Sep/2015
+
+References
+----------
+
+http://quant-econ.net/jl/ddp.html
+
+Notes
+-----
+1. This currently implements ... Value Iteration, Policy Iteration, Modified Policy Iteration
+2. This does not currently support state-action pair formulation, or sparse matrices
+
+=#
+import Base.*
+
+#-----------------#
+#-Data Structures-#
+#-----------------#
+
+"""
+DiscreteDP type for specifying paramters for discrete dynamic programming model
+
+##### Parameters
+
+- `R::Array{T,NR}` : Reward Array
+- `Q::Array{T,NQ}` : Transition Probability Array
+- `beta::Float64`  : Discount Factor
+
+##### Returns
+
+- `ddp::DiscreteDP` : DiscreteDP object
+
+"""
+type DiscreteDP{T<:Real,NQ,NR}
+    R::Array{T,NR}  #-Reward Array-#
+    Q::Array{T,NQ}  #-Transition Probability Array-#
+    beta::Float64   #-Discount Factor-#
+
+    function DiscreteDP(R::Array, Q::Array, beta::Float64)
+        # verify input integrity
+        !(NQ in [2, 3]) && throw(ArgumentError("Q must be 2- or 3-dimensional"))
+        !(NR in [1, 2]) && throw(ArgumentError("R must be 1- or 2-dimensional"))
+        beta < 0 || beta >= 1 &&  throw(ArgumentError("beta must be [0, 1)"))
+
+        # check feasibility
+        R_max = s_wise_max(R)
+        if any(R_max .== -Inf)
+            # First state index such that all actions yield -Inf
+            s = find(R_max .== -Inf) #-Only Gives True
+            throw(ArgumentError("for every state the reward must be finite for
+                some action: violated for state $s"))
+        end
+
+        new(R, Q, beta)
+    end
+end
+
+# necessary for dispatch to fill in type parameters {T,NQ,NR}
+DiscreteDP{T,NQ,NR}(R::Array{T,NR}, Q::Array{T,NQ}, beta::Float64) =
+    DiscreteDP{T,NQ,NR}(R, Q, beta)
+
+#~Property Functions~#
+num_states(ddp::DiscreteDP) = size(ddp.R, 1)
+num_actions(ddp::DiscreteDP) = size(ddp.R, 2)
+
+abstract DDPAlgorithm
+immutable VFI <: DDPAlgorithm end
+immutable PFI <: DDPAlgorithm end
+immutable MPFI <: DDPAlgorithm end
+
+"""
+DPSolveResult is an object for retaining results and associated metadata after
+solving the model
+
+##### Parameters
+
+- `ddp::DiscreteDP` : DiscreteDP object
+
+##### Returns
+
+- `ddpr::DPSolveResult` : DiscreteDP Results object
+
+"""
+type DPSolveResult{Algo<:DDPAlgorithm}
+    v::Vector{Float64}
+    Tv::Array{Float64}
+    num_iter::Int
+    sigma::Array{Int,1}
+    mc::MarkovChain
+
+    function DPSolveResult(ddp::DiscreteDP)
+        v = s_wise_max(ddp.R)                         #Initialise v with v_init
+        ddpr = new(v, similar(v), 0, similar(v, Int))
+
+        # fill in sigma with proper values
+        compute_greedy!(ddp, ddpr)
+        ddpr
+    end
+
+    # method to pass initial value function (skip the s-wise max)
+    function DPSolveResult(ddp::DiscreteDP, v::Vector)
+        ddpr = new(v, similar(v), 0, similar(v, Int))
+
+        # fill in sigma with proper values
+        compute_greedy!(ddp, ddpr)
+        ddpr
+    end
+end
+
+#-----------------#
+#-Inner Functions-#
+#-----------------#
+
+"""
+Return the vector max_a vals(s, a), where vals is represented
+by a 2-dimensional ndarray of shape (self.num_states,
+self.num_actions). Stored in out, which must be of length
+self.num_states.
+out and out_argmax must be of length self.num_states; dtype of
+out_argmax must be int.
+
+"""
+s_wise_max(vals::Matrix) = vec(maximum(vals, 2))
+s_wise_max!(vals::Matrix, out::Vector) = maximum!(out, vals)
+s_wise_max!{T}(vals::Matrix{T}, out::Vector, out_argmax) =
+    Base.findminmax!(Base.MoreFun(), fill!(out, -Inf) , out_argmax, vals)
+
+#=
+Define Matrix Multiplication between 3-dimensional matrix and a vector
+
+Matrix multiplication over the last dimension of A
+
+=#
+function *{T}(A::Array{T,3}, v::Vector)
+    size(A, 3) ==  size(v, 1) || error("wrong dimensions")
+    out = Array(T, size(A)[1:2])
+
+    # TODO: slicing A[i, j, :] is cache-unfriendly
+    for j=1:size(out, 2), i=1:size(out, 1)
+        out[i, j] = dot(A[i, j, :][:], v)
+    end
+    out
+end
+
+"""
+The Bellman operator, which computes and returns the updated value function Tv
+for a value function v.
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the model parameters
+- `ddpr::DPSolveResult` : Object that contains result variables
+
+##### Returns
+
+- `ddpr.Tv::Array{Float64,1}` : Updated value function vector
+
+##### Notes
+
+Updates `ddpr.Tv` and `ddpr.sigma` inplace
+
+"""
+function bellman_operator!(ddp::DiscreteDP, ddpr::DPSolveResult)
+    vals = ddp.R + ddp.beta * ddp.Q * ddpr.v
+    s_wise_max!(vals, ddpr.Tv, ddpr.sigma)
+    ddpr.Tv
+end
+
+"""
+The Bellman operator, which computes and returns the updated value function Tv
+for a given value function v.
+
+This function will fill the input `v` with `Tv` and the input `sigma` with the
+corresponding policy rule
+
+##### Parameters
+
+- `ddp::DisreteDP`: The ddp model
+- `v::Vector{Float64}`: The current guess of the value function
+- `sigma::Vector{Int}`: A buffer array to hold the policy function. Initial
+  values not used and will be overwritten
+
+##### Returns
+
+- `Tv::Vector{Float64}`: Updated value function vector
+- `sigma::Vector{Float64}`: Policy rule
+"""
+function bellman_operator!(ddp::DiscreteDP, v::Vector{Float64}, sigma::Vector{Int})
+    vals = ddp.R + ddp.beta * ddp.Q * v
+    s_wise_max!(vals, v, sigma)
+    v, sigma
+end
+
+"""
+The Bellman operator, which computes and returns the updated value function Tv
+for a given value function v.
+
+##### Parameters
+
+- `ddp::DisreteDP`: The ddp model
+- `v::Vector{Float64}`: The current guess of the value function
+
+##### Returns
+
+- `Tv::Array{Float64,1}` : Updated value function vector
+"""
+bellman_operator(ddp::DiscreteDP, v::Vector{Float64}) =
+    s_wise_max(ddp.R + ddp.beta * ddp.Q * v)
+
+
+"""
+Compute the v-greedy policy
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the model parameters
+- `ddpr::DPSolveResult` : Object that contains result variables
+
+##### Returns
+
+- `sigma::Array{Int,1}` : Array containing `v`-greedy policy rule
+
+##### Notes
+
+modifies ddpr.sigma and ddpr.Tv in place
+
+"""
+compute_greedy!(ddp::DiscreteDP, ddpr::DPSolveResult) =
+    (bellman_operator!(ddp, ddpr); ddpr.sigma)
+
+function compute_greedy(ddp::DiscreteDP, v::Array{Float64})
+    Tv = copy(v)  # buffer so we don't change the input v
+    sigma = ones(Int, length(v))
+    bellman_operator!(ddp, Tv, sigma)
+    sigma
+end
+
+#-Policy Iteration-#
+
+"""
+Method of `evaluate_policy` that extracts sigma from a `DPSolveResult`
+
+See other docstring for details
+"""
+evaluate_policy(ddp::DiscreteDP, ddpr::DPSolveResult) =
+    evaluate_policy(ddp, ddpr.sigma)
+
+"""
+Compute the value of a policy.
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the model parameters
+- `sigma::Array{Int}` : Policy rule vector
+
+##### Returns
+
+- `v_sigma::Array{Float64}` : Value vector of `sigma`, of length n.
+
+"""
+function evaluate_policy(ddp::DiscreteDP, sigma::Vector{Int})
+    R_sigma, Q_sigma = RQ_sigma(ddp, sigma)
+    b = R_sigma
+    A = I - ddp.beta * Q_sigma
+    return A \ b
+end
+
+"""
+Solve the dynamic programming problem.
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the Model Parameters
+- `method::Type{T<Algo}(VFI)`: Type name specifying solution method. Acceptable
+arguments are `VFI` for value function iteration or `PFI` for policy function
+iteration
+- `;max_iter::Int(250)` : Maximum number of iterations
+- `;epsilon::Float64(1e-3)` : Value for epsilon-optimality. Only used if
+`method` is `VFI`
+- `;k::Int(20)` : Number of iterations for partial policy evaluation in modified
+policy iteration (irrelevant for other methods).
+
+##### Returns
+
+- `ddpr::DPSolveResult{Algo}` : Optimization result represented as a
+DPSolveResult. See `DPSolveResult` for details.
+"""
+function solve{Algo<:DDPAlgorithm}(ddp::DiscreteDP, method::Type{Algo}=VFI;
+                                   max_iter::Int=250, epsilon::Float64=1e-3,
+                                   k::Int=20)
+    ddpr = DPSolveResult{Algo}(ddp)
+    _solve!(ddp, ddpr, max_iter, epsilon, k)
+    ddpr.mc = MarkovChain(ddp, ddpr)
+    ddpr
+end
+
+function solve{Algo<:DDPAlgorithm}(ddp::DiscreteDP, v_init::Vector,
+                                   method::Type{Algo}=VFI; max_iter::Int=250,
+                                   epsilon::Float64=1e-3, k::Int=20)
+    ddpr = DPSolveResult{Algo}(ddp, v_init)
+    _solve!(ddp, ddpr, max_iter, epsilon, k)
+    ddpr.mc = MarkovChain(ddp, ddpr)
+    ddpr
+end
+
+Base.ind2sub(ddp::DiscreteDP, ddpr::DPSolveResult) =
+    map(x->ind2sub(size(ddp.R), x)[2], ddpr.sigma)
+
+"""
+Impliments Value Iteration
+NOTE: See `solve` for further details
+"""
+function _solve!(ddp::DiscreteDP, ddpr::DPSolveResult{VFI}, max_iter::Int64,
+               epsilon::Float64, k::Int)
+    if ddp.beta == 0.0
+        tol = Inf
+    else
+        tol = epsilon * (1-ddp.beta) / (2*ddp.beta)
+    end
+
+    for i in 1:max_iter
+        # updates Tv in place
+        bellman_operator!(ddp, ddpr)
+
+        # compute error and update the v inside ddpr
+        err = maxabs(ddpr.Tv .- ddpr.v)
+        copy!(ddpr.v, ddpr.Tv)
+        ddpr.num_iter += 1
+
+        if err < tol
+            break
+        end
+    end
+
+    ddpr
+end
+
+"""
+Policy Function Iteration
+
+NOTE: The epsilon is ignored in this method. It is only here so dispatch can
+      go from `solve(::DiscreteDP, ::Type{Algo})` to any of the algorithms.
+      See `solve` for further details
+"""
+function _solve!(ddp::DiscreteDP, ddpr::DPSolveResult{PFI}, max_iter::Int,
+                epsilon::Float64, k::Int)
+    old_sigma = copy(ddpr.sigma)
+
+    for i in 1:max_iter
+       ddpr.v = evaluate_policy(ddp, ddpr)
+       compute_greedy!(ddp, ddpr)
+
+       if all(old_sigma .== ddpr.sigma)
+           break
+       end
+       copy!(old_sigma, ddpr.sigma)
+       ddpr.num_iter += 1
+    end
+
+    ddpr
+end
+
+span(x::Vector) = maximum(x) - minimum(x)
+midrange(x::Vector) = mean(extrema(x))
+
+"""
+Modified Policy Function Iteration
+
+NOTE: The epsilon is ignored in this method. It is only here so dispatch can
+      go from `solve(::DiscreteDP, ::Type{Algo})` to any of the algorithms.
+      See `solve` for further details
+"""
+function _solve!(ddp::DiscreteDP, ddpr::DPSolveResult{MPFI}, max_iter::Int,
+                 epsilon::Float64, k::Int)
+    beta = ddp.beta
+    fill!(ddpr.v, minimum(ddp.R[ddp.R .> -Inf]) / (1.0 - beta))
+    old_sigma = copy(ddpr.sigma)
+
+    tol = beta > 0 ? epsilon * (1-beta) / beta : Inf
+
+    for i in 1:max_iter
+        bellman_operator!(ddp, ddpr)  # updates Tv, sigma inplace
+        dif = ddpr.Tv - ddpr.v
+
+        # check convergence
+        if span(dif) < tol
+            ddpr.v = ddpr.Tv + midrange(dif) * beta / (1-beta)
+            break
+        end
+
+        # now update v to use the output of the bellman step when entering
+        # policy loop
+        copy!(ddpr.v, ddpr.Tv)
+
+        # now do k iterations of policy iteration
+        R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
+        for i in 1:k
+            ddpr.Tv = R_sigma + beta * Q_sigma * ddpr.v
+
+            # compute error and update the v inside ddpr
+            err = maxabs(ddpr.Tv .- ddpr.v)
+            copy!(ddpr.v, ddpr.Tv)
+        end
+
+        ddpr.num_iter += 1
+
+    end
+
+    ddpr
+end
+
+#----------------------------#
+#-Suppoting Markov Functions-#
+#----------------------------#
+
+"""
+Returns the controlled Markov chain for a given policy `sigma`.
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the model parameters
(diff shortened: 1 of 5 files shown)


spencerlyon2:
Before we merge let's move the notebooks out of the `src` directory

spencerlyon2:
@oyamad I have a question for you here.

`s_wise_max` is well suited for numpy's row major semantics, but less cache friendly for Julia's column major ordering. Do you see any reason not to "transpose" everything so `s_wise_max` can take max over columns instead of rows?

It would diverge from the python version, but maybe this is a worthwhile divergence?

src/markov/ddp.jl
```diff
@@ -0,0 +1,374 @@
...
+    A = I - ddp.beta * Q_sigma
+    return A \ b
+end
+
+# TODO: should we just inline this inside solve{PFI}?
+
+
+#-----------------#
+#-Outer Functions-#
+#-----------------#
```
spencerlyon2: This comment doesn't really apply to this code in the same way as the python

src/markov/ddp.jl
```diff
@@ -0,0 +1,374 @@
...
+    tol = epsilon * (1-ddp.beta) / (2*ddp.beta)
+
+    # perform VFI
+    for i in 1:max_iter
+        # updates Tv in place
+        bellman_operator!(ddp, ddpr)
+
+        # compute error and update the v inside ddpr
+        err = maxabs(ddpr.Tv .- ddpr.v)
+        copy!(ddpr.v, ddpr.Tv)           #Equivalent to `v[:] =` in numpy
```
spencerlyon2: Remove intermediate comments we inserted when mapping from numpy to julia

src/markov/ddp.jl
```diff
@@ -0,0 +1,374 @@
...
+        copy!(ddpr.v, ddpr.Tv)
+
+        # now do k iterations of policy iteration
+        for i in 1:k
+            R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
+            ddpr.Tv = R_sigma + beta * Q_sigma * ddpr.v
+
+            # compute error and update the v inside ddpr
+            err = maxabs(ddpr.Tv .- ddpr.v)
+            copy!(ddpr.v, ddpr.Tv)           #Equivalent to `v[:] =` in numpy
```
spencerlyon2: Remove the comment

test/test_discrete_dp.jl
```diff
@@ -0,0 +1,136 @@
...
+
+# Example from Puterman 2005, Section 3.1
+beta = 0.95
+
+# Formulation with R: n x m, Q: n x m x n
+n, m = 2, 2  # number of states, number of actions
+R = [5 10; -1 -Inf]
+Q = Array(Float64, n, m, n)
+#-TODO:Check this indexing-#
+Q[0, 0, :] = 0.5, 0.5
```
spencerlyon2: Can't index with 0...

src/markov/ddp.jl
mmcky: addressed.

src/markov/ddp.jl
mmcky: addressed.

mmcky:
I am going through the ``src/markov/ddp.jl`` file now updating all of the function docstrings.

oyamad:
@mmcky @spencerlyon2 This is great!

@spencerlyon2 Regarding transposing everything, we should do so if it leads to speedup. But I guess we might wait to implement the state-action pairs formulation to see what the best solution will be in both formulations.

src/markov/ddp.jl
```diff
@@ -0,0 +1,394 @@
...
+    end
+
+    ddpr
+end
+
+span(x::Vector) = maximum(x) - minimum(x)
+midrange(x::Vector) = mean(extrema(x))
+
+"""
+Policy Function Iteration
```
oyamad: Modified Policy Function Iteration

oyamad:
It seems argmax is not properly obtained. What does `Base.findminmax!` on line 127 do?

oyamad:
See [this notebook](http://nbviewer.ipython.org/gist/oyamad/6faca3a2d65ae00ed64a), where I ran the code with the simple example in Puterman. The returned policy function is `[1 2]`, but action `2` at state `2` cannot be optimal, which is infeasible and assigned a value `-Inf`.

src/markov/ddp.jl
```diff
@@ -0,0 +1,394 @@
...
+        # now do k iterations of policy iteration
+        for i in 1:k
+            R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
+            ddpr.Tv = R_sigma + beta * Q_sigma * ddpr.v
+
+            # compute error and update the v inside ddpr
+            err = maxabs(ddpr.Tv .- ddpr.v)
+            copy!(ddpr.v, ddpr.Tv)
+
+            if err < tol
```
oyamad: This part is not there in the Python version.

There is a tradeoff between the benefit of avoiding "unnecessary" iterations by breaking when `err < tol` and the cost of computing `err = maxabs(ddpr.Tv .- ddpr.v)` in every iteration, and this depends on the value of `tol`. In general, there is no reason that the tolerance here should be the same value as the tolerance used in the main loop for  `span(dif) < tol`. If we keep this part, maybe better to give the option of choose an independent tolerance value?

src/markov/ddp.jl
```diff
@@ -0,0 +1,394 @@
...
+            break
+        end
+
+        # now update v to use the output of the bellman step when entering
+        # policy loop
+        copy!(ddpr.v, ddpr.Tv)
+
+        # now do k iterations of policy iteration
+        for i in 1:k
+            R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
```
oyamad: This does not need to be computed inside the `for i in 1:k` loop.

Would it be possible to implement the `T_sigma` operation as in the [Python version](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/ddp.py#L513-L529)?

src/markov/ddp.jl
```diff
@@ -0,0 +1,394 @@
...
+- `Q::Array{T,NQ}` : Transition Probability Array
+- `R::Array{T,NR}` : Reward Array
+- `beta::Float64`  : Discount Factor
+
+##### Returns
+
+- `ddp::DiscreteDP` : DiscreteDP object
+
+"""
+type DiscreteDP{T<:Real,NQ,NR}
```
oyamad: `DiscreteDP(Q, R, beta)` versus `DiscreteDP(R, Q, beta)`: wouldn't it be better to have a common order between the Julia and Python versions?

src/markov/ddp.jl
```diff
@@ -0,0 +1,394 @@
...
+    ddpr
+end
+
+"""
+Impliments Value Iteration
+NOTE: See `solve` for further details
+"""
+function _solve!(ddp::DiscreteDP, ddpr::DPSolveResult{VFI}, max_iter::Int64,
+               epsilon::Float64, k::Int)
+    tol = epsilon * (1-ddp.beta) / (2*ddp.beta)
```
oyamad: The case where `beta = 0` should be handled separately.

oyamad:
Would it be possible to implement a version of `bellman_operator` with which we can write the code like:

```julia
for i in 1:num_iter
    v_new = bellman_operator(ddp, v)
    # Print or plot v for exmple here
    copy!(v, v_new)
end
```

(I know this is possible with `bellman_operator!(ddp, ddpr)` by defining `ddpr`, but the above seems more intuitive.)

spencerlyon2:
Hey @oyamad thanks for all the comments, I'll do my best to answer all of them:

- `Base.findminmax!(Base.Functor, out_val , out_argval, input_matrix)` will fill `out_val` and `out_argval` with the minimum/argmin or `max/argmax`  along the singleton dimensions of `out_val` and `out_argval`. Note that `out_argval` will be filled with a **linear index** into `input_matrix`, not an index column by column. I'm pretty confident that it is doing this correctly. To get the column by column maximizer can [do this](https://github.com/QuantEcon/QuantEcon.jl/pull/78/files?diff=unified#diff-96652b67c79d6c06d266bb1ae747d05eR388). If this still doesn't look right to you let me know. Also, if we don't like this behavior of returning a linear index, we can just write our own implementation. 
- [ ] We can definitely switch the order of input args to match python
- [ ] We can implement the `bellman_operator(::DDP, ::Array) -> Array`
- [ ] Good call about handling beta=0 case. 
- [ ] We should take the call to `RQ_sigma` out of the `1:k` loop
-  I'm happy to get rid of the tolerance check on the inside of the `1:k` loop.

oyamad:
@spencerlyon2 Thank you for your reply.

I am not sure if I really understand how to manipulate `ddpr`, but `compute_greedy!` seems not working properly.

```julia
julia> include("ddp.jl")
DDP

julia> n, m = 2, 2;

julia> R = [5 10; -1 -Inf]
2x2 Array{Float64,2}:
  5.0    10.0
 -1.0  -Inf  

julia> Q = Array(Float64, n, m, n);

julia> Q[1, 1, :] = [0.5, 0.5];

julia> Q[1, 2, :] = [0, 1];

julia> Q[2, 1, :] = [0, 1];

julia> Q[2, 2, :] = [0.5, 0.5]  # Arbitrary;

julia> beta = 0.95;

julia> ddp = DDP.DiscreteDP(Q, R, beta)
DDP.DiscreteDP{Float64,3,2}(2x2x2 Array{Float64,3}:
[:, :, 1] =
 0.5  0.0
 0.0  0.5

[:, :, 2] =
 0.5  1.0
 1.0  0.5,2x2 Array{Float64,2}:
  5.0    10.0
 -1.0  -Inf  ,0.95)

julia> ddpr = DDP.DPSolveResult{DDP.VFI}(ddp)
DDP.DPSolveResult{DDP.VFI}([10.0,-1.0],[9.275,-1.95],0,[1,2],#undef)

julia> ddpr.v
2-element Array{Float64,1}:
 10.0
 -1.0

julia> ddpr.sigma
2-element Array{Int64,1}:
 1
 2

julia> ddpr.v = [0; 0]
2-element Array{Float64,1}:
 0.0
 0.0

julia> DDP.compute_greedy!(ddp, ddpr)
2-element Array{Int64,1}:
 3
 2

julia> ddpr.sigma
2-element Array{Int64,1}:
 3
 2

```

Given `v = [0; 0]`, the `v`-greedy policy should be `[2; 1]`. `3` is out of range anyway...

oyamad:
The problem seems to be in `s_wise_max!`:

```julia
julia> vals = [5 10; -1 -Inf]
2x2 Array{Float64,2}:
  5.0    10.0
 -1.0  -Inf  

julia> Tv = Vector{Float64}(2);

julia> sigma = Vector{Int64}(2);

julia> DDP.s_wise_max!(vals, Tv, sigma)
([10.0,-1.0],[3,2])

julia> sigma
2-element Array{Int64,1}:
 3
 2

```

oyamad:
Sorry, now I think I understand what you mean by "linear order":

```julia
julia> ind = map(x->ind2sub(size(vals), x), sigma)
2-element Array{Tuple{Int64,Int64},1}:
 (1,2)
 (2,1)

```

oyamad:
I think `bellman_operator(:: DiscreteDP, ::Array) -> Array` looks more straightforward than `(:: DiscreteDP, :: DPSolveResult)`, so it may be good to have the former *in addition to* (not instead of) the latter.

The user may use `bellman_operator` in a loop for example to see what is going on inside `solve` or solve a finite-horizon DP.

spencerlyon2:
@oyamad that's great, so do you think that `s_wise_max!` and the indices are correct now?

And yeah, I agree with that comment about the `bellman_operator` method. We should implement that function also

oyamad:
> @oyamad that's great, so do you think that s_wise_max! and the indices are correct now?

Yes, sorry for my confusion!

spencerlyon2:
OK great and no problem. I don't love the behavior either -- it isn't very intuitive to me

oyamad:
I translated `random_discrete_dp` into Julia in [`random_ddp`](https://github.com/QuantEcon/QuantEcon.jl/tree/random_ddp) branch.

Usage:

```julia
julia> num_states, num_actions = 3, 2;

julia> beta = 0.95;

julia> ddp = random_discrete_dp(num_states, num_actions, beta)
QuantEcon.DDP.DiscreteDP{Float64,3,2}(3x2x3 Array{Float64,3}:
[:, :, 1] =
 0.237836  0.360343 
 0.575364  0.664304 
 0.575823  0.0909946

[:, :, 2] =
 0.156995  0.387196
 0.109598  0.314492
 0.091807  0.89442 

[:, :, 3] =
 0.605169  0.252461 
 0.315039  0.0212035
 0.33237   0.0145855,3x2 Array{Float64,2}:
 -1.20203   -0.661307
 -0.715116  -0.349859
  0.968529  -1.66845 ,0.95)

julia> res_pi = DDP.solve(ddp, DDP.PFI);

julia> res_pi.num_iter
1

julia> res_pi.sigma
3-element Array{Int64,1}:
 1
 2
 3

julia> res_pi.v
3-element Array{Float64,1}:
 -4.19214
 -4.18071
 -2.46888

julia> res_vi = DDP.solve(ddp, DDP.VFI);

julia> res_vi.num_iter
173

julia> res_vi.sigma
3-element Array{Int64,1}:
 1
 2
 3

julia> res_vi.v
3-element Array{Float64,1}:
 -4.19164
 -4.18021
 -2.46839

julia> res_mpi = DDP.solve(ddp, DDP.MPFI);

julia> res_mpi.num_iter
2

julia> res_mpi.sigma
3-element Array{Int64,1}:
 1
 2
 3

julia> res_mpi.v
3-element Array{Float64,1}:
 -4.19214
 -4.18071
 -2.46888

```

I hope it helps...

spencerlyon2:
Yes, that's great. It will definitely help. Thanks

mmcky:
@spencerlyon2 I just pushed some updates to address the check boxes. How can I push a removal of the last commit? `8606075`. It has the correct interface but isn't right as ``bellman_operator`` sits within a loop. Do we need to pass ``Tv`` and ``sigma`` for the ``s_wise_max!`` operation?

spencerlyon2:
@mmcky thanks for working on this.

Stack overflow suggests [this](http://stackoverflow.com/questions/927358/how-do-you-undo-the-last-commit?page=1&tab=active#tab-top) for undoing a commit

The answer to your question about `s_wise_max!` will depend on context and which version of the method you are talking about.

mmcky:
@spencerlyon2 Just getting back to this now. I looked into the workflow and decided to run a ``git revert``. That way it adds a commit to undo the changes in a specific commit and it doesn't break anyone else's work.

mmcky:
Regarding new bellman_operator for use within loops. 
Is this the function we are looking for, passing in ``v::Array`` and ``sigma::Array``?

```julia
function bellman_operator(ddp::DiscreteDP, v::Array{Float64}, sigma::Array{Int,1})
    vals = ddp.R + ddp.beta * ddp.Q * v
    Tv = similar(v)
    s_wise_max!(vals, Tv, sigma)
    return Tv
end
```

oyamad:
1. Is it possible to implement both `bellman_operator(ddp, v)` which only returns `Tv` and `bellman_operator!(ddp, v, sigma)` which overwrites `sigma` with the `v`-greedy policy?

2. It will be nice if the user can easily write code of finite-horizon backward induction as in [this notebook](http://nbviewer.ipython.org/github/oyamad/mdp/blob/master/ddp_ex_MF_7_6_4.ipynb), cell `[9]`.

3. This is not necessary, but it would also be nice if we have `evaluate_policy(sigma)` and `compute_greedy(v)`, in addition to `evaluate_policy(ddp, ddpr)` and `compute_greedy!(ddp, ddpr)`, to see how the algorithms work by implementing them using these functions, as in [this notebook](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/ddp/ddp_theory.ipynb), cells `[11]` and `[23]`.

jstac:
@oyamad Just a side comment:  These are beautiful notebooks.  Next week my brother, who is a web developer, will arrive here and we'll do some work on the top level of the lecture site.  One thing we want to do is give prominence to a curated list of notebooks, kind of like [this page](https://github.com/ipython/ipython/wiki/A-gallery-of-interesting-IPython-Notebooks).  We'll take out the ones now in QuantEcon.site and put all notebooks we like in [this repo](https://github.com/QuantEcon/QuantEcon.notebooks).  If you're willing, it would be great to have the full collection of your DP notebooks.

oyamad:
@jstac Sure, please feel free to pick any of the notebooks there!

(One concern is, as reported [here](https://github.com/QuantEcon/QuantEcon.py/issues/185#issuecomment-153020375), that the figure at the bottom (right panel) in [this notebook](http://nbviewer.ipython.org/github/oyamad/mdp/blob/master/ddp_ex_MF_7_6_4.ipynb) looks different from the original figure in Miranda and Fackler.)

spencerlyon2:
OK, I think we have taken care of all the requests/concerns in this PR.

I'll wait for the green light from Travis and give others a day to comment before merging.

src/markov/ddp.jl
```diff
@@ -0,0 +1,468 @@
...
+
+    for i in 1:max_iter
+       ddpr.v = evaluate_policy(ddp, ddpr)
+       compute_greedy!(ddp, ddpr)
+
+       if all(old_sigma .== ddpr.sigma)
+           break
+       end
+       copy!(old_sigma, ddpr.sigma)
+       ddpr.num_iter += 1
```
oyamad: This should come before the `if` block? (Or should initialize `ddpr` with `num_iter=1`?)

src/markov/ddp.jl
```diff
@@ -0,0 +1,468 @@
...
+        R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
+        for i in 1:k
+            ddpr.Tv = R_sigma + beta * Q_sigma * ddpr.v
+
+            # compute error and update the v inside ddpr
+            err = maxabs(ddpr.Tv .- ddpr.v)
+            copy!(ddpr.v, ddpr.Tv)
+        end
+
+        ddpr.num_iter += 1
```
oyamad: The same applies here.

src/markov/ddp.jl
```diff
@@ -0,0 +1,468 @@
...
+        # policy loop
+        copy!(ddpr.v, ddpr.Tv)
+
+        # now do k iterations of policy iteration
+        R_sigma, Q_sigma = RQ_sigma(ddp, ddpr)
+        for i in 1:k
+            ddpr.Tv = R_sigma + beta * Q_sigma * ddpr.v
+
+            # compute error and update the v inside ddpr
+            err = maxabs(ddpr.Tv .- ddpr.v)
```
oyamad: This is no longer necessary.

src/markov/ddp.jl
```diff
@@ -0,0 +1,468 @@
...
+
+##### Parameters
+
+- `ddp::DisreteDP` : Object that contains the Model Parameters
+- `method::Type{T<Algo}(VFI)`: Type name specifying solution method. Acceptable
+arguments are `VFI` for value function iteration or `PFI` for policy function
+iteration
+- `;max_iter::Int(250)` : Maximum number of iterations
+- `;epsilon::Float64(1e-3)` : Value for epsilon-optimality. Only used if
+`method` is `VFI`
```
oyamad: Also used in `MPFI`.

src/markov/ddp.jl
```diff
@@ -0,0 +1,468 @@
+#=
+Discrete Decision Processes
+
+@author : Daisuke Oyama (Ported by Spencer Lyon, Matthew McKay)
```
oyamad: Three of us (or just two of you guys) are the authors.

oyamad:
The code looks very good. The only thing is that I am still a bit concerned about the "linear order" approach. It may be just a matter of taste though.

oyamad:
I updated `random_discrete_dp`, which is now in [`random_ddp2 `](https://github.com/QuantEcon/QuantEcon.jl/tree/random_ddp2) branch.

spencerlyon2:
The linear order is definitely not as intuitive as the row by row index. 

It is, however,  commonly used in julia especially in the functions from Base that we are using here. 

We did add one helper method to `Base.ind2sub` that takes a `ddp` and a `sigma` and gives back the row based index version of sigma. 

Another option would be to have another field on the result type that has type `Nullable{Vector {Int}}` that we could leave blank at first  (to conserve memory) and have a special method that would populate it with the row based sigma if the user asks us to. Would that make it easier to reason about?

oyamad:
@spencerlyon2 Thanks, fair enough.

> Another option would be to have another field on the result type that has type Nullable{Vector {Int}} that we could leave blank at first (to conserve memory) and have a special method that would populate it with the row based sigma if the user asks us to. Would that make it easier to reason about?

Now I do not think this is necessary.

In many cases, what the user is interested in is not the index itself (whether linear or row based), but the action value,  for example, the level of consumption or saving associated with each action index. This is related to the discussion in QuantEcon/QuantEcon.py#101.

I realized that for the state-action pair formulation, `a_indices` can actually be used to store action values (not action indices).

src/markov/ddp.jl
```diff
@@ -0,0 +1,469 @@
...
+    ddpr
+end
+
+span(x::Vector) = maximum(x) - minimum(x)
+midrange(x::Vector) = mean(extrema(x))
+
+"""
+Modified Policy Function Iteration
+
+NOTE: The epsilon is ignored in this method. It is only here so dispatch can
```
oyamad: `epsilon` is used in MPFI.

src/markov/ddp.jl
```diff
@@ -0,0 +1,469 @@
...
+    end
+end
+
+#-----------------#
+#-Inner Functions-#
+#-----------------#
+
+"""
+Return the vector max_a vals(s, a), where vals is represented
+by a 2-dimensional ndarray of shape (self.num_states,
```
oyamad: Should be translated to Julian terminologies.

src/markov/ddp.jl
```diff
@@ -0,0 +1,469 @@
...
+
+- `ddp::DisreteDP`: The ddp model
+- `v::Vector{Float64}`: The current guess of the value function
+- `sigma::Vector{Int}`: A buffer array to hold the policy function. Initial
+  values not used and will be overwritten
+
+##### Returns
+
+- `Tv::Vector{Float64}`: Updated value function vector
+- `sigma::Vector{Float64}`: Policy rule
```
oyamad: Should be `Vector{Int}`.

src/markov/ddp.jl
```diff
@@ -0,0 +1,469 @@
...
+"""
+function bellman_operator!(ddp::DiscreteDP, v::Vector{Float64}, sigma::Vector{Int})
```
oyamad: When I tried, `v = [0, 0, 0]` wasn't accepted, but had to do `v = [0., 0., 0.]`. Would it be possible to allow `v` to be any `Real`?

spencerlyon2:
@oyamad thank you for continuing to be so thorough reviewing this PR.

I believe I have addressed all the points you brought up and made relevant changes in the code

src/markov/ddp.jl
```diff
@@ -0,0 +1,472 @@
...
+"""
+function bellman_operator!{T<:AbstractFloat}(ddp::DiscreteDP, v::Vector{T},
+                                             sigma::Vector)
+    vals = ddp.R + ddp.beta * ddp.Q * v
+    s_wise_max!(vals, v, sigma)
+    v, sigma
+end
+
+# method to allow users to pass v as something like v=[1, 1, 1]
+bellman_operator!{T<:Real}(ddp::DiscreteDP, v::Vector{T}, sigma::Vector) =
```
oyamad: Is it possible to make it

* `bellman_operator!(ddp, v, Tv, sigma)`: input `ddp` and `v` and overwrite `Tv` and `sigma` (and do not modify `v`)

or

* `bellman_operator!(ddp, v, sigma)`: input `ddp` and `v`, return `Tv`, and overwrite `sigma` (and do not modify `v`)

Sorry for so many requests..

oyamad: Sorry, the above comment is for line 197.

oyamad: Just to note, for this one `bellman_operator!{T<:Real}(ddp::DiscreteDP, v::Vector{T}, sigma::Vector)`, the input `v` seems not to be modified.

src/markov/ddp.jl
```diff
@@ -0,0 +1,472 @@
...
+
+##### Returns
+
+- `ddp::DiscreteDP` : DiscreteDP object
+
+"""
+type DiscreteDP{T<:Real,NQ,NR}
+    R::Array{T,NR}  #-Reward Array-#
+    Q::Array{T,NQ}  #-Transition Probability Array-#
+    beta::Float64   #-Discount Factor-#
```
oyamad: Should this be `Real` as well?

src/markov/ddp.jl
```diff
@@ -0,0 +1,472 @@
...
+
+##### Returns
+
+- `R_sigma::Array{Float64}`: Reward vector for `sigma`, of length n.
+
+- `Q_sigma::Array{Float64}`: Transition probability matrix for `sigma`,
+  of shape (n, n).
+
+"""
+function RQ_sigma{T<:Integer}(ddp::DiscreteDP, sigma::Array{T})
```
oyamad: This may be worth exporting?

oyamad:
I translated [this notebook](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/ddp/ddp_theory.ipynb) in Python into Julia [here](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/ddp/ddp_theory_jl.ipynb). The examples from Puterman are correctly reproduced.

(I do not know how to manipulate the `DataFrame` tables (the column names and the row indices). Could anyone kindly edit them...)

oyamad:
@spencerlyon2 I have a few more questions:

0. [Maybe for future enhancement] If the elements of `R` and `Q` as well as `beta` are allowed to be `Rational`, will the outputs be `Rational` with exact arithmetics (in particular for `PFI`)? It is the case for `MarkovChain`.

0. You once mentioned transposing everything, which is something that has to be done at some point, I suppose. What would that be like? I just wondered if the "linear order" in the current implementation would be preserved there.

src/markov/ddp.jl
spencerlyon2: No problem about the requests. We can add these methods. Do you worry that the number of API points is getting overwhelming?

And yes, that last method is a bit odd. I'll re-work it. We will lose some efficiency, but it will be the correct behavior.

spencerlyon2: Ok so more comments about the last method. There are times when `v` cannot be modified in place.

For example, suppose that you pass in a `Vector{Int}` and you end up getting that the result of the bellman operator is not integer-valued (say, a `Vector{Float64}`). When you try to update the values of the input `v` with the result of the bellman operation you will get an `InexactError` because the new `Float64` values can't be represented as `Int`.

So, we have four options:

1. Remove this method and force users to pass in floating point numbers (just doing `[1, 1, 1.0]` or `Float64[1, 1, 1]` instead of `[1,1,1]` is sufficient)
2. Keep the method as it is and don't update `v`
3. Alter the method to try to mutate the input and just let the error happen
4. Alter the method to try to mutate, catch it if it happens, and give a more detailed error message when re-throwing the error.

Which alternative do you prefer?

src/markov/ddp.jl
spencerlyon2: Done

spencerlyon2:
Hi @oyamad I think I have addressed the latest set of comments.

There was one unresolved line comment that disappeared when I altered the diff, so I'll post the relevant part of my comment here again.

You asked about the method `bellman_operator!{T<:Real}(ddp::DiscreteDP, v::Vector{T}, sigma::Vector)` not updating `v` in place. Here was my response:

---

Ok so more comments about the last method. There are times when v cannot be modified in place.

For example, suppose that you pass in a Vector{Int} and you end up getting that the result of the bellman operator is not integer-valued (say, a Vector{Float64}). When you try to update the values of the input v with the result of the bellman operation you will get an InexactError because the new Float64 values can't be represented as Int.

So, we have four options:

1. Remove this method and force users to pass in floating point numbers (just doing [1, 1, 1.0] or Float64[1, 1, 1] instead of [1,1,1] is sufficient)
2. Keep the method as it is and don't update v
3. Alter the method to try to mutate the input and just let the error happen
4. Alter the method to try to mutate, catch it if it happens, and give a more detailed error message when re-throwing the error.

Which alternative do you prefer?

oyamad:
@spencerlyon2 Thank you for your response.


0. My previous comment (followed by "Sorry, the above comment is for line 197") was about the original one `bellman_operator!{T<:AbstractFloat}(ddp::DiscreteDP, v::Vector{T}, sigma::Vector)` with `T<:AbstractFloat`, where I would like the input `v` to be kept untouched, not replaced with `Tv`. So the two candidates:
   * `bellman_operator!(ddp, v, Tv, sigma)`: input `ddp` and `v` and overwrite `Tv` and `sigma` (and do not modify `v`)

   or

   * `bellman_operator!(ddp, v, sigma)`: input `ddp` and `v`, return `Tv`, and overwrite `sigma` (and do not modify `v`)

   Do you think making one of these would be sensible?

0. Regarding `bellman_operator!{T<:Real}(ddp::DiscreteDP, v::Vector{T}, sigma::Vector)` with `T<:Real`, option 1 (remove this) will make things simplest (yes, [1, 1, 1.0] or Float64[1, 1, 1] instead of [1,1,1] is sufficient).

oyamad:
Sorry, it's already there! (Please forget the first one.)

spencerlyon2:
Ok sounds good. 

We are getting close. Do you mind taking another look when you have a chance to give the (final ?) approval?

oyamad:
I think this is ready to be merged. Great work, thanks!

(I updated the [demonstration notebook](http://nbviewer.ipython.org/github/QuantEcon/QuantEcon.site/blob/mdp/_static/notebooks/inwork/ddp/ddp_theory_jl.ipynb).)

spencerlyon2:
That notebook is great, thanks for putting it together.

I'm glad we could get the code into the right form to be useful to you.

I'm curious what the performance is relative to the python version... I wouldn't be surprised if it was a bit slower here because we have the problem set up in a cache un-friendly way

jstac:
Great work guys, thanks all!!

@oyamad We're working on getting all the Jupyter notebooks up in a revamped site over the next week or two (while Andrij is visiting us here at NYU).  Would you mind to put all or any notebooks you'd like to contribute into [the notebooks repo](https://github.com/QuantEcon/QuantEcon.notebooks)?  All the notebooks of yours I've seen are great so we'll put them all up.  

The [notebooks](https://github.com/QuantEcon/QuantEcon.site/tree/master/_static/notebooks) in QuantEcon.site we'll shift over ourselves, probably later today.

oyamad:
@jstac Please pick any of them. Will you put the notebooks yourself? Or should I make a pull request to the notebooks repo?

jstac:
@oyamad A PR would be great.  Like I said, as many of your notebooks you'd like to post we'd like to have up there.

oyamad:
Sure (I will do tomorrow morning).

oyamad:
Some [performance tests](http://nbviewer.ipython.org/gist/oyamad/c161449d73a620d53ff8) using `random_discrete_dp`. Compare the [Python version](http://nbviewer.ipython.org/github/oyamad/mdp/blob/master/ddp_performance.ipynb). Timing technology is not the same across the two, but the Julia version is sometimes 10x slower...

spencerlyon2:
Thanks for posting. We have to remember that we should let the jit run on the Julia code and not do timing on the first call.

spencerlyon2:
Ahh nevermind. I see that you have done this. Disregard the previous comment

mmcky:
@oyamad @spencerlyon2 Thanks for the help in getting this merged.

PR #80, opened 2015-11-11
BUG: Fix bug in probvec

oyamad:
For `k=1`, `probvec` (and hence `random_stochastic_matrix` and `random_markov_chain`) raises an error with version 0.4, due to the change in the behavior of `sort`.

Version 0.4.1:

```julia
julia> k, m = 1, 3;

julia> r = rand(k-1, m)
0x3 Array{Float64,2}

julia> sort(r, 1)
ERROR: ArgumentError: step cannot be zero
 in steprange_last at /Applications/Julia-0.4.1.app/Contents/Resources/julia/lib/julia/sys.dylib
 in sort_chunks! at sort.jl:495
 in sort at sort.jl:489

```

Version 0.3.11:

```julia
julia> k, m = 1, 3;

julia> r = rand(k-1, m)
0x3 Array{Float64,2}

julia> sort(r, 1)
0x3 Array{Float64,2}

```

Fix by handling the case of `k=1` separately.

### DIFF
--- a/src/random_mc.jl
+++ b/src/random_mc.jl
@@ -164,6 +164,9 @@ Return m randomly sampled probability vectors of size k.
 
 """
 function random_probvec(k::Integer, m::Integer)
+    k == 1 && return ones((k, m))
+
+    # if k >= 2
     x = Array(Float64, (k, m))
 
     r = rand(k-1, m)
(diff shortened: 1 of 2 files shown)


spencerlyon2:
Thanks!

PR #85, opened 2015-12-24
ddp: Re-implement 3-dim matrix * vector

oyamad:
I was reading [this article](http://julialang.org/blog/2013/09/fast-numeric/), and applied the tips of "Write cache-friendly codes" there to `*` in `ddp.jl`.

* This operation itself became 10x faster: see [this notebook](http://nbviewer.jupyter.org/gist/oyamad/75c9630268278d8d54c8).

* Overall `solve` became 2x-3x faster:
  * [current version](http://nbviewer.jupyter.org/gist/oyamad/c161449d73a620d53ff8/random_ddp_jl01.ipynb)
  * [this version](http://nbviewer.jupyter.org/gist/oyamad/c161449d73a620d53ff8/random_ddp_jl02.ipynb)

I am not sure if such a local optimization as this makes much sense, but open this PR for discussion.

### DIFF
--- a/src/markov/ddp.jl
+++ b/src/markov/ddp.jl
@@ -388,14 +388,19 @@ Matrix multiplication over the last dimension of A
 
 =#
 function *{T}(A::Array{T,3}, v::Vector)
-    size(A, 3) ==  size(v, 1) || error("wrong dimensions")
-    out = Array(T, size(A)[1:2])
+    n1, n2, n3 = size(A)
+    size(v, 1) == n3 || error("wrong dimensions")
+    out = Array(T, n1, n2)
+    N = n1 * n2
 
-    # TODO: slicing A[i, j, :] is cache-unfriendly
-    for j=1:size(out, 2), i=1:size(out, 1)
-        out[i, j] = dot(A[i, j, :][:], v)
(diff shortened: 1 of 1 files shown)


oyamad:
`OverflowError()` is raised for `Rational` with version 0.5. I wonder what is going on...

spencerlyon2:
This is great, thanks for taking care of that todo item!

The overflow is expected. I'm assuming you have numbers of type `Rational{Int64}`, which means the numerator and denominator both have to be signed 64 bit integers. I'm guessing some of the computations would violate this constraint so instead of returning something imprecise, they throw this error. For true exact computation with rationals use `Rational {BigInt}`. I had to do this in the tests . See these two parts of the tests:

https://github.com/QuantEcon/QuantEcon.jl/blob/master/test/test_ddp.jl#L66

https://github.com/QuantEcon/QuantEcon.jl/blob/master/test/test_ddp.jl#L141

oyamad:
@spencerlyon2 Thanks for the comments.

But I still wonder why the test passes with 0.4 but doesn't with 0.5 ("nightly")?

oyamad:
Replaced by #86.

PR #86, opened 2015-12-27
ddp: Re-implement 3-dim matrix * vector with reshape

oyamad:
Replaces #85. This is even faster (except for small-size arrays) and simpler.
* [`reduce_last2`](http://nbviewer.jupyter.org/gist/oyamad/75c9630268278d8d54c8/reduce_last_jl02.ipynb)

The speed of `solve` is more or less the same or slightly faster.
* [this version](http://nbviewer.jupyter.org/gist/oyamad/c161449d73a620d53ff8/random_ddp_jl03.ipynb)

### DIFF
--- a/src/markov/ddp.jl
+++ b/src/markov/ddp.jl
@@ -388,14 +388,13 @@ Matrix multiplication over the last dimension of A
 
 =#
 function *{T}(A::Array{T,3}, v::Vector)
-    size(A, 3) ==  size(v, 1) || error("wrong dimensions")
-    out = Array(T, size(A)[1:2])
+    shape = size(A)
+    size(v, 1) == shape[end] || error("wrong dimensions")
 
-    # TODO: slicing A[i, j, :] is cache-unfriendly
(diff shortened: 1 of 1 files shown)


oyamad:
Another test failure: "test methods for subtypes != (Float64, Int)".

`@fact f(ddp0, [1.0, 1.0]) --> f(ddp0, ones(T, 2))`: Do they need to match exactly?

spencerlyon2:
I don't really see why this shouldn't be exact, but I think it is definitely ok to throw a `roughly` in here.

spencerlyon2:
And great work on this new method. I wouldn't have guessed that `reshape` would have been faster than rolling out the loops, but it looks like it is for larger arrays. I'm in favor of this once we sort out tests and get travis to give us a green light here.

spencerlyon2:
Ping.

I'll merge once tests are passing. I'm happy to lossen testing requirements to make comparisons not exact.

oyamad:
@spencerlyon2 Thank you!

PR #88, opened 2016-01-04
Change type in meshgrid function

cc7768:
Changes the types of the function arguments in `meshgrid` from `Vector` to `AbstractVector`.


spencerlyon2:
Tests _should_ be passing.

If so, this closes #86

PR #91, opened 2016-01-10
use BetaBinomial from Distributions.jl instead of QuantEcon Fixes #67

### LINKED ISSUE #67: ENH: move BetaBinomial to Distributions.jl
We really don't need to define this distribution here.

It would be far better to have this in Distributions.jl

Before this can be down we need to fill in [these methods](http://distributionsjl.readthedocs.org/en/latest/univariate.html#common-interface). 

It is highly likely that we only need to define a subset of them and the rest will be filled in via dispatch on an abstract type.

We should consult the source of Distributions.jl and construct a list of which methods we actually need to implement ourselves before submitting the PR.

rawls238:
@spencerlyon2


spencerlyon2:
Thanks for doing this. Good work working with the Distributions.jl team to get it merged in

The doc files are actually autogenerated from our source code, so next time I run the docgen script those changes will get overwriten. That's ok though, thanks for working on this.

PR #92, opened 2016-01-11
update tests to use BaseTestNext instead of FactCheck Fixes #89

### LINKED ISSUE #89: Move to new Base.Test framework
Right now we are using `FactCheck.jl` mainly to group tests.

On 0.5-dev we can now use `@testset` to accomplish this.

On 0.4 we can use the `BaseTestNext.jl` library to use the new 0.5 `Base.Test` module.

See [here](https://github.com/EconForge/NLsolve.jl/blob/master/test/runtests.jl#L3-L8) for an example of how to do this

rawls238:
@spencerlyon2

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -2,5 +2,5 @@ QuantEcon
 HDF5
 MAT
 JLD
-FactCheck
 DataStructures
+BaseTestNext
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -1,5 +1,7 @@
 using Compat  # for startswith
 using QuantEcon
+using DataStructures: counter
+using Distributions: LogNormal, pdf
 
 if VERSION >= v"0.5-"
     using Base.Test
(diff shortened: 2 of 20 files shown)


rawls238:
@spencerlyon2 this PR does a number of things mainly moves from FactCheck -> use the v0.5 Base.Test methods, removes the use of modules from the Test packages to replace copied imports in each of the projects, opting to use modules only when it requires explicit external exports that other packages don't require, and cleans up a few things in `runtests` that were preventing me from running tests locally though seemed to be working in travis.

spencerlyon2:
Looks great overall, thanks for doing this work.

I like the change to remove modules and would suggest that we do so entirely. 

I believe we need 

- `DataStructures.counter` in `test_discrete_rv.jl` 
- `Distributions.LogNormal` `Distributions.pdf` in `test_lae.jl`

I would recommend just putting:

```julia
using DataStructures: counter
using Distributions: LogNormal, pdf
```

at the top of runtests.jl

rawls238:
sounds good to me, will make those changes tonight

rawls238:
@spencerlyon2 should be good to go

spencerlyon2:
Looks great, thanks

PR #93, opened 2016-01-19
don't need to include compat in tests anymore

rawls238:
@spencerlyon2 this isn't needed anymore in the tests as it isn't used


spencerlyon2:
Sounds good, thanks!

PR #94, opened 2016-01-23
Adding State-Action Pair Support

albep:
Looking forward to your comments and suggestions on my implementation in Julia of the state-action pair formulation from the Python codes


oyamad:
@albep Thanks! I will take a closer look as soon as possible.

Just one quick question: Is it possible to unify `DiscreteDP` and `DiscreteDP_sa`? If you have seen difficulties in doing this, can you share them with us?

If unified, `DiscreteDP` would have `s_indices::Union{Void,Vector{Tind}}` and so on with `s_indices=nothing`. Would it be inefficient? SA type would be differentiated with `DiscreteDP{T,2,1,Tbeta,Tind}` from `DiscreteDP{T,3,2,Tbeta,Tind}`.

albep:
@oyamad Thank you for your comments! Yes, I believe that can be done but hadn't thought about differentiating methods for `s_wise_max` based on that type formulation. Someone more qualified than me should speak about the relative efficiency of the two formulations, but I think yours would be preferable as it only relies on concrete types. @spencerlyon2 ?

spencerlyon2:
I'm going to review this this morning.

@oyamad makes a great point. The reason we have so many type parameters was precisely so that we could leave room in the original type for the SA pairs formulation. 

I think that we could do something like this:

```
typealias DDP{T,Tbeta,Tind} DiscreteDP{T,2,1,Tbeta,Tind}
typealias DDPsa{T,Tbeta,Tind} DiscreteDP{T,3,2,Tbeta,Tind}
```

and then use `::DiscreteDP` on methods that work the same for both types, `::DDP` for ones that use the current formulation and `::DDPsa` on  methods that apply to the s-a formulation.

I'm happy to change the names of those type alias, but I do think that defining them would be a good way forward.

spencerlyon2:
I think that instead of 

```julia
s_indices::Union{Void,Vector{Tind}}
```

We would do 

```julia
s_indices::Nullable{Vector{Tind}}
```

which will avoid type stability issues. The reason is that if we have the type on the field be an explicit union, the compiler will always have to treat it as such and do dynamic type lookup every time we use it. However, if we use `Nullable`, the concrete type of `s_indices` will always be `Nullable{Vector{Tind}}` -- which completely avoids the type inference/stability problem. 

When we actually use the `s_indices ` in the `DDPsa` algorithms we will simply have a call 

```julia
s_indices = get(ddp. s_indices)
```

which will also be type stable because at this point the compiler knows for sure that `s_indices` will be a `Vector{Tind}`

spencerlyon2:
OK I took a few minutes to sort out the types here. I've created a new branch [refactor-ddp](https://github.com/QuantEcon/QuantEcon.jl/tree/refactor-ddp) that adds one commit on top of this PR that implements the ideas suggested above.

@albep here's my proposal for getting this thing merged:

1. You  can take over and continue working on top of the new `refactor-ddp` branch. Make sure you are happy with how the types were implemented.
2. Add some tests so that we can make sure this is working. We should be able to borrow from the [python tests](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/tests/test_ddp.py#L32-L44)
3. Open a new pull request that uses the unified type and has some tests
4. I'll then review that and we can iterate until the code is ready to be included. 

If we are happy with that plan, I think we can close this pull request for right now.

oyamad:
@spencerlyon2 I didn't know of `Nullable`. Good information.

I have some concern about the "number of actions" (for both the py and jl versions).

1.
First of all, `num_actions` is (almost) not used, so the issue here is not of practical importance, but just conceptual.

2.
There are two approaches to define the set(s) of actions.

1. For each state `s`, defined `A(s)` to be the set of actions available at `s`, and think of `A(s)` and `A(s')` as totally different (i,e., they are disjoint) for `s != s'`. That is, consider actions under different states as different even if they are attached the same label. In this case, the number of actions is equal to the number of state-action pairs.

2. Define `A` to be the universal set of actions, and consider the set `A(s)` of actions available at `s` as a subset of `A`. The union of all `A(s)`s may be a proper subset of `A`. In this case, the number of actions is either the number of elements of `A` or the number of elements of the union of `A(s)`s.

I think approach 1 is legitimate, but approach 2 may be more natural in some cases.

3.
In the Python version, I followed approach 2, where with the SA formulation, it is defined by `self.num_actions = self.a_indices.max() + 1` (in Julia it should be `num_actions(ddp) = maximum(ddp. a_indices)`) which I now find feels a bit strange.

For example, if `s_indices = [1, 1, 2]` and `a_indices = [1, 2, 100]`, then the number of actions will be 100. This is totally fine, but this information is almost useless.

The bottom line is: I propose to delete `num_actions` both in the julia and in the python versions. We can just do`n, m = size(R)` etc.

albep:
@spencerlyon2 Thanks a lot for the changes you've made. I'm happy with them. I also like your plan and will start working down those points.

@oyamad I noticed that `num_actions` is rarely used, in fact when translating your Python code for the state-action pair formulation into Julia I left it out completely, as it seems it's never used when we have state-action pairs.

If we agree, I can delete `num_action` from the Julia version when I get to work on it again.

spencerlyon2:
Merged in #100

PR #99, opened 2016-03-14
Refactor markov

spencerlyon2:
This implements a few things:

- reorganize markov chain tools into the `markov` folder
- loosen some type constraints to be `AbstractArray` instead of only `Array`
- Add support for sparse transition matrices in the `MarkovChain` type.


src/markov/mc_tools.jl
```diff
@@ -400,7 +365,7 @@ initial states.
 """
 function simulate!(mc::MarkovChain, X::Matrix{Int})
     n = size(mc.p, 1)
-    P_dist = [DiscreteRV(vec(mc.p[i, :])) for i in 1:n]
+    P_dist = [DiscreteRV(vec(full(mc.p[i, :]))) for i in 1:n]
```
spencerlyon2: This is the one main thing I don't feel great about in this PR. Notice that I added `full` around each row of the stochastic matrix. 

`DiscreteRV` uses a `cumsum` over an `AbstractVector` to give us the transition to the next state. In Julia 0.4 there is no `SparseVector` type (only sparse matrix), so to get this to be an `AbstractVector` we needed to make it dense before passing to `DiscreteRV`. The thing that makes it ok is that `cumsum` is an operation that will make any sparse array dense, after the first non-zero element. So we probably are more efficient working directly with the dense array than we would have been working with a dense sparse array, but I just wanted to point this out.

Also, I checked the native machine code for this function with a dense `mc.p` before and after making this change and the machine code was identical. this means that when `mc.p` is already dense the jit compiler completely removes the no-op that `full` causes and we have 0 performance hit.

spencerlyon2:
Note, travis failures here are only on julia 0.5 (downstream issue with a dependency not loading on unreleased julia). 

I have updated the travis.yml file in #100 to allow failures on 0.5, but haven't duplicated that commit here. If there are no objections I'm ok to merge this.

spencerlyon2:
Added a few more things:

- `state_values` as a field on `MarkovChain`. This is any `AbstractVector` and defaults to `1:size(p, 1)`, where `p` is the transition matrix
- Methods `simulate_values` to do a simulation and report `state_values` instead of state indices
- `simulate` now always returns a Matrix (as does `simulate_values`)
- To do a single simulation and get a vector use `simulation` or `value_simulation`
- Extended `DiscreteRV` to allow for `q` and `Q` to be different types of `AbstractVector`. I did this so `q` could be a `SubArray` in `simulate!` and `simulate_values!` to avoid allocations
- Added inner constructor check on `DiscreteRV` that `sum(q) == Q[end] == 1.0`

spencerlyon2:
failing tests are a result of a dependency not working on the dev version of Julia. In another PR I have specified that these tests are allowed to fail. We can merge this without waiting for tests to pass.

spencerlyon2:
Last change for now, I promise: `tauchen` and `rouwenhorst` now return instances of `MarkovChain` instead of a state_values, p tuple

oyamad:
This doesn't have to belong to this PR, but `random_markov_chain` and `random_stochastic_matrix` in [random_mc.jl](https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/random_mc.jl) should support sparse matrix (by the way, this file should go to `markov/` too). Compare [the python version](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/random.py#L111-L147).

spencerlyon2:
Good point, see #102

Issue #103, opened 2016-03-17
Drop `num_actions` from DiscreteDP


oyamad:
As in the Python version QuantEcon/QuantEcon.py#236.

oyamad:
Done by #207

PR #110, opened 2016-05-14
Modify src/QuantEcon.jl

myuuuuun:
* delete evaluate_policy!  : which doesn't exist in ddp.jl
* export num_states  : for convenience


spencerlyon2:
Thanks for cleaning this up @myuuuuun1

@oyamad are you ok with this? For some reason I'm remembering you commenting about the future of `num_states` or `num_actions`.

oyamad:
It is `num_actions` that I propose to delete (#103), while I believe `num_states` is worth exporting.

spencerlyon2:
Excellent, thanks for confirming

PR #116, opened 2016-07-08
Fixing bug related to w_path and some related comments

albep:
I fixed the type inconsistency and I also changed a few words in the comments at the end to be consistent with the fact that `w_path`, in the original code (not following my changes), is not a vector of `N(0,1)` innovations but it's stored as `C*N(0,1)` vector. This is still somewhat different than how it's described in the initial documentation, as there it's said that `w` is `N(0,1)`

### DIFF
--- a/src/lqcontrol.jl
+++ b/src/lqcontrol.jl
@@ -276,23 +276,23 @@ function _compute_sequence{T}(lq::LQ, x0::Vector{T}, policies)
 
     x_path = Array(T, n, capT+1)
     u_path = Array(T, k, capT)
-    w_path = [vec(C*randn(j)) for i=1:(capT+1)]
+    w_path = C*randn(j, capT+1)
 
     x_path[:, 1] = x0
     u_path[:, 1] = -(first(policies)*x0)
 
     for t = 2:capT
         f = policies[t]
-        x_path[:, t] = A*x_path[: ,t-1] + B*u_path[:, t-1] + w_path[t]
+        x_path[:, t] = A*x_path[: ,t-1] + B*u_path[:, t-1] + w_path[:, t]
         u_path[:, t] = -(f*x_path[:, t])
     end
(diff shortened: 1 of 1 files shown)


src/lqcontrol.jl
```diff
@@ -308,8 +308,8 @@ only for `min(ts_length, lq.capT)`
 represents `x_t`
 - `u_path::Matrix{Float64}` : A k x T matrix, where the t-th column represents
 `u_t`
-- `w_path::Matrix{Float64}` : A j x T+1 matrix, where the t-th column represents
-`lq.C*w_t`
+- `w_path::Matrix{Float64}` : A n x T+1 matrix, where the t-th column represents
+`lq.C*N(0,1)`
```
spencerlyon2: Was the old doc string incorrect here? 

I can't recall, but I believe that putting `w_t` at the end should be right. If so, isn't that more specific than `N(0,1)` (i.e. We are saying which standard normal RV we have, not just an arbitrary one)

albep: I thought that looked better because at the beginning we say that `w_t` is a white noise, but then `w_path` - which is supposed to hold the entire history of `w_t`'s - actually holds `C * w_t`, so at that point it seemed better to write `C * N(0,1)`. Anyways, I'm happy with reverting that to the original

spencerlyon2: Let's get another opinion on this. 

@QuantEcon/lead-developers can anyone comment on this?

jstac:
I like @albep 's change -- as he says, the notation is a bit inconsistent and this seems like a good workaround.

albep:
Thanks @jstac. Let me know if there's any other change you'd like being made to this code. Otherwise, once merged, I'll turn to fixing the notebook in `QuantEcon.applications` https://github.com/QuantEcon/QuantEcon.applications/issues/10

Issue #117, opened 2016-07-09
DiscreteDP: Should not modify s_indices and a_indices when sorting


oyamad:
```jl
julia> using QuantEcon

julia> s_indices = [2, 1, 1];  # Not sorted;

julia> a_indices = [1, 1, 2];

julia> R = [-1., 5, 10];

julia> Q = [0 1; 0.5 0.5; 0 1];

julia> beta = 0.95;

julia> ddp = DiscreteDP(R, Q, beta, s_indices, a_indices);

julia> s_indices
3-element Array{Int64,1}:
 1
 1
 2

julia> a_indices
3-element Array{Int64,1}:
 1
 1
 2
```

One cannot reuse `s_indices` and `a_indices`:

```jl
julia> ddp = DiscreteDP(R, Q, beta, s_indices, a_indices);
ERROR: ArgumentError: Duplicate s-a pair found
 in anonymous at /Users/oyama/.julia/v0.4/QuantEcon/src/markov/ddp.jl:131
 in sparse at sparse/csparse.jl:93
 in call at /Users/oyama/.julia/v0.4/QuantEcon/src/markov/ddp.jl:130
 in call at /Users/oyama/.julia/v0.4/QuantEcon/src/markov/ddp.jl:205
```

spencerlyon2:
cc @albep

albep:
@oyamad thanks for pointing that out. PR #122 fixes that, I tried it with your example

PR #122, opened 2016-07-11
Fix indices overwr ddp

albep:
This commit fixes the overwriting of `s_indices` (and `a_indices`) when sorting, which prevented re-use of the same vectors. See #117


spencerlyon2:
@albep codecov seems to suggest that the patch you made here makes some sections not covered by tests. Can you look into this?

oyamad:
In fact, `s_indices` is not necessary to include as a field: it is never used once `a_indptr` is constructed.

(We should remove `self.s_indices` in the Python version as well.)

albep:
Thanks  @oyamad for pointing that out. The last commit addresses that issue as well.

PR #123, opened 2016-07-12
Fixed Matrix Multiplication in DDP

MaximilianJHuber:
Matrix-vector multiplication is carried out first, instead of matrix-matrix multiplication. This cuts the number of memory allocations and runtime.


spencerlyon2:
Great work, thanks!

PR #126, opened 2016-07-15
Use Primes package

davidanthoff:
This fixes a precompile error on julia 0.5 and should also work on 0.4.


spencerlyon2:
Thanks!

PR #128, opened 2016-07-15
bug fix in Markov chain simulation

amckay:
With num_reps > 1 the t+1 draw of the chain was always using the distribution conditional on the value at t for the first repetition.

### DIFF
--- a/src/markov/mc_tools.jl
+++ b/src/markov/mc_tools.jl
@@ -382,7 +382,7 @@ function simulate!(mc::MarkovChain, X::Matrix{Int})
 
     for i in 1:k
         for t in 1:ts_length-1
-            X[t+1, i] = draw(P_dist[X[t]])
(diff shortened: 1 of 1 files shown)


oyamad:
@amckay Thanks! On `master` we have

```jl
julia> using QuantEcon

julia> mc = MarkovChain([0 1; 1 0]);

julia> X = Array(Int, 5, 2);

julia> X[1, :] = [1, 2];

julia> simulate!(mc, X)
5x2 Array{Int64,2}:
 1  2
 2  2
 1  1
 2  2
 1  1
```

This is definitely a bug, and it is fixed by this PR.

jstac:
@amckay Many thanks!  
@spencerlyon2 This is ready to merge.

spencerlyon2:
Great, thank you!

PR #129, opened 2016-07-16
add minimum version requirement for Compat.view

tkelman:
also add StatsBase explicitly to REQUIRE, on the off-chance it's ever not available
as an indirect dependency


spencerlyon2:
Thanks

PR #130, opened 2016-07-18
Upgraded code to handle 1 dimensional cases

albep:
When fixing the notebook `lqcontrol_solutions_jl` I encountered a situation where matrix `Z` on this line https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/matrix_eqn.jl#L116 is actually an `Array{Float64, 1}` with a single element (as `R` is a scalar). As `Base.cond` requires a matrix, this commit makes `Z` into a comformable `Array{Float64, 2}` always.

### DIFF
--- a/src/matrix_eqn.jl
+++ b/src/matrix_eqn.jl
@@ -113,7 +113,7 @@ function solve_discrete_riccati(A::ScalarOrArray, B::ScalarOrArray,
     BTA = B' * A
 
     for gamma in candidates
-        Z = reshape(R + gamma .* BB, n, n)
+        Z = getZ(R, gamma, BB)
         cn = cond(Z)
         if isfinite(cn)
             Q_tilde = -Q + N' * (Z \ (N + gamma .* BTA)) + gamma .* I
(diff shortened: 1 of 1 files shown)


spencerlyon2:
Test is failing... https://travis-ci.org/QuantEcon/QuantEcon.jl/jobs/145585643#L341

albep:
Yes, that's because the test is running a completely scalar case, and `reshape(Float, n, n)` is not defined. Hadn't thought about that. 

Let me describe the issue. We have this multiplication `Z = R + gamma .* (B' * B)`, and then `cn = cond(Z)`. `B` can be a matrix or a vector. In the first case, no problem. In the second case, as long as `B` is `Array{Float, 1}`, `B' * B` is `Array{Float, 1}`. So, when passing `Z` to `cond()`, we get an error as `cond()` wants `Array{Float, 2}` or `Float`. At this point, do you think it's better to just leave the code as it is and letting the user figure out that he must write the vector `B` as `Array{Float, 2}`?

src/matrix_eqn.jl
```diff
@@ -168,3 +168,48 @@ function solve_discrete_riccati(A::ScalarOrArray, B::ScalarOrArray,
...
+
+- `BB::Array{Float64, 1}` : result of B' * B
+- `gamma::Float64` : parameter in the Riccati equation solver
+- `R::Float64`
+
+##### Returns
+- `::Float64` : element Z in the Riccati equation solver
+
+"""
+getZ(R::Float64, gamma::Float64, BB::Array{Float64, 1}) = R + gamma * BB[1]
```
spencerlyon2: This is really an internal method, so it probably isn't a concerns, but I am a little worried about assuming BB has only one element in it.  Should we add a check here?

albep: We could, but I think there is no way it has more than 1 element, as it comes from `B' * B` - where `B` is either a `Vector` or `Matrix`

albep: By the way, why is the test failing? I run the tests on my computer and they worked

spencerlyon2:
The failure isn't our fault: https://travis-ci.org/QuantEcon/QuantEcon.jl/jobs/145860837#L291

LightGraphs.jl isn't compiling

albep:
I thought my code might have been responsible for LightGraphs.jl not compiling. Let me know if you'd like to make any other changes before merging

PR #133, opened 2016-07-31
Use dirname(@__FILE__) instead of Pkg.dir

tkelman:
This allows installing the package elsewhere.

Add testing against 0.5 to Travis - this runs the most
recent RC now, release once final tags are done

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,17 @@ language: julia
 sudo: false
 julia:
     - 0.4
+    - 0.5
     - nightly
 matrix:
   allow_failures:
     - julia: nightly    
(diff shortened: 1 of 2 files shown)


spencerlyon2:
Thanks for re-opening. It looks like the test failure is now actually our fault where before it was due to a dependency failing to install. Just for personal record I think I'll keep this open as a reminder to fix that issue, unless anyone has objections.

tkelman:
that's fine, no rush on this - was just having trouble loading the old travis log, wanted to poke it again

spencerlyon2:
Good to go now, thanks.

Issue #119, opened 2016-07-10
MarkovChain: state_values related


oyamad:
1. Change
   - (`simulate`, `simulate_values`) to (`simulate_indices`, `simulate`)
   - (`simulation`, `value_simulation`) to (`simulation_indices`, `simulation`)?
2. For the new `simulate` (current `simulate_values`), `init` should be specified as a state value, not index. Need a method to return the index in `state_values` given a state value (for the Python version, see https://github.com/QuantEcon/QuantEcon.py/pull/237/commits/5bc78d5706462cde82dcba09dfec114c2a3e4680).
3. Let `recurrent_classes` and `communication_classes` return state values, and add `*_indices`.

oyamad:
`simulate_values!(mc::MarkovChain, init_state::Vector{Int}, X::Matrix)` currently works as follows:

```jl
julia> mc = MarkovChain([0 1; 1 0]);

julia> X = Array(Int, 5, 2);

julia> init_state = [1, 2];

julia> simulate_values!(mc, init_state, X)
5x2 Array{Int64,2}:
 4448600528  4448600944
          2           1
          1           2
          2           1
          1           2
```

It should be better to fill `X[1, :]` with the initial state values.

jstac:
@oyamad I'm confused.  What's the difference between `simulation` and `simulate`?   It seems confusing to have both for instances of `MarkovChain`.

oyamad:
`simulation` returns a `Vector`, while `simulate` returns a `Matrix` (even when num_reps=1) (https://github.com/QuantEcon/QuantEcon.jl/pull/99#issuecomment-197412344). I agree it can be confusing, but I have no good idea for alternatives.

jstac:
Thanks.  I guess I knew that and forgot.  Let's open this up to discussion.

First, I like @oyamad 's plan to unify names with what we have in Python.  The Python conventions are simple and clear.  In Python we only have `simulate` and `simulate_indices`, each of which generate a single time series.  So my preference is:

1. Change (`simulate`, `simulate_values`) to (`simulate_indices`, `simulate`) and have them return flat arrays.

2. Drop `simulation` and `value_simulation`.

oyamad:
[...]

PR #137, opened 2016-08-18
Redesign of the MarkovChain simulation methods

vgregory757:
In reference to #119. There are now four simulation methods that do the following:

* `simulate`: simulates one sample path of the Markov chain and returns a `Vector`
* `simulate!`: fills a pre-allocated `Matrix` with one simulation of the chain down each column
* `simulate_indices` and `simulate_indices!`: same as above, except they return sample paths with the indices of the state values as elements

I implemented @spencerlyon2's idea about initial conditions. The initial conditions can either be blank, a scalar, or a vector, with implications for each described in the docstrings or [here](https://github.com/QuantEcon/QuantEcon.jl/issues/119#issuecomment-238163616).

### DIFF
--- a/src/markov/mc_tools.jl
+++ b/src/markov/mc_tools.jl
@@ -320,10 +320,10 @@ The resulting matrix has the state values of `mc` as elements.
 
 ### Arguments
 
-- `mc::MarkovChain` : MarkovChain instance.
 - `X::Matrix` : Preallocated matrix to be filled with sample paths
 of the Markov chain `mc`. The element types in `X` should be the 
 same as the type of the state values of `mc`
+- `mc::MarkovChain` : MarkovChain instance.
 - `;init=rand(1:n_states(mc))` : Can be one of the following
     - blank: random initial condition for each chain
     - scalar: same initial condition for each chain
@@ -405,9 +405,9 @@ The resulting matrix has the indices of the state values of `mc` as elements.
 
 ### Arguments
 
-- `mc::MarkovChain` : MarkovChain instance.
 - `X::Matrix{Int}` : Preallocated matrix to be filled with indices
 of the sample paths of the Markov chain `mc`.
+- `mc::MarkovChain` : MarkovChain instance.
 - `;init=rand(1:n_states(mc))` : Can be one of the following
     - blank: random initial condition for each chain
     - scalar: same initial condition for each chain
--- a/src/markov/mc_tools.jl
+++ b/src/markov/mc_tools.jl
@@ -328,11 +328,13 @@ same as the type of the state values of `mc`
     - blank: random initial condition for each chain
     - scalar: same initial condition for each chain
     - vector: cycle through the elements, applying each as an
-      initial condition
+      initial condition until all columns have an initial condition 
+      (allows for more columns than initial conditions)
 """
 
-function simulate!(mc::MarkovChain, X::Matrix;
-                   init=rand(1:n_states(mc), size(X, 2)))
+function simulate!(X::Union{AbstractVector,AbstractMatrix},
+                   mc::MarkovChain; init=rand(1:n_states(mc), size(X, 2)))
+
     # Note: eltype(X) must be the same as eltype(mc.state_values)
     if eltype(X) != eltype(mc.state_values)
         throw(ArgumentError("Types in X must be the same as the types of the
@@ -343,13 +345,12 @@ function simulate!(mc::MarkovChain, X::Matrix;
     
     # if init is a vector, assign initial conditions to columns of X
     # otherwise, just start each column at the same initial condition
-    real_init = Array(Int, k, )
     if length(init) > 1
         real_init = collect(take(cycle(init), k))
     else
-        real_init = round(Int, init.*ones(k, ))
+        real_init = init .* ones(Int, k)
     end
-    X[1, :] = mc.state_values[real_init]'
+    X[1, :] = mc.state_values[real_init]
 
     n = size(mc.p, 1)
 
@@ -410,23 +411,23 @@ of the sample paths of the Markov chain `mc`.
     - blank: random initial condition for each chain
     - scalar: same initial condition for each chain
     - vector: cycle through the elements, applying each as an
-      initial condition
+      initial condition until all columns have an initial condition 
+      (allows for more columns than initial conditions)
 """
 
-function simulate_indices!(mc::MarkovChain, X::Matrix{Int};
-                           init=rand(1:n_states(mc), size(X, 2)))
+function simulate_indices!{T<:Integer}(X::Union{AbstractVector{T},AbstractMatrix{T}},
+                           mc::MarkovChain; init=rand(1:n_states(mc), size(X, 2)))
 
     ts_length, k = size(X)
    
     # if init is a vector, assign initial conditions to columns of X
     # otherwise, just start each column at the same initial condition
-    real_init = Array(Int, k, )
     if length(init) > 1
         real_init = collect(take(cycle(init), k))
     else
-        real_init = round(Int, init.*ones(k, ))
+        real_init = init .* ones(Int, k)
     end
-    X[1, :] = real_init'
+    X[1, :] = real_init
 
     n = size(mc.p, 1)
 
--- a/src/markov/mc_tools.jl
+++ b/src/markov/mc_tools.jl
@@ -291,86 +291,66 @@ todense(::Type, A::Array) = A
 
 
 """
-Simulate time series of state transitions of the Markov chain `mc`.
+Simulate one sample path of the Markov chain `mc`.
+The resulting vector has the state values of `mc` as elements.
 
-The sample path from the `j`-th repetition of the simulation with initial state
-`init[i]` is stored in the `(j-1)*num_reps+i`-th column of the matrix X.
-
-##### Arguments
+### Arguments
 
 - `mc::MarkovChain` : MarkovChain instance.
-- `ts_length::Int` : Length of each simulation.
-- `init::Vector{Int}` : Vector containing initial states.
-- `;num_reps::Int(1)` : Number of repetitions of simulation for each element
-of `init`.
+- `ts_length::Int` : Length of simulation
+- `;init::Int=rand(1:n_states(mc))` : Initial state 
 
-##### Returns
+### Returns
 
-- `X::Matrix{Int}` : Array containing the sample paths as columns, of shape
-(ts_length, k), where k = length(init)* num_reps.
-
+- `X::Vector` : Vector containing the sample path, with length 
+ts_length
 """
-function simulate(mc::MarkovChain, ts_length::Int, init::Vector{Int};
-                  num_reps::Int=1)
-    k = length(init)*num_reps
-    X = Array(Int, ts_length, k)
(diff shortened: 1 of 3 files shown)


src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
 
-- `mc::MarkovChain` : MarkovChain instance.
-- `X::Matrix{Int}` : Preallocated matrix of integers to be filled with sample
-paths of the markov chain `mc`. The elements in `X[1, :]` will be used as the
-initial states.
+    ts_length, k = size(X)
+    
+    # if init is a vector, assign initial conditions to columns of X
+    # otherwise, just start each column at the same initial condition
+    real_init = Array(Int, k, )
```
cc7768: Not sure we need these trailing commas in the array declarations (there are a couple other instances of this below). It leaves things as a vector, but seems to be extraneous code -- Happy to hear if others disagree, but I would vote for their removal.

jstac:
@vgregory757 Many thanks for working on this.  The failing test is in `test_mc_tools.jl`.  I guess it needs to be updated.

I don't know what the rules are with trailing commas in `Array()` calls.

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
+    if length(init) > 1
+        real_init = collect(take(cycle(init), k))
```
cc7768: Is there a reason that we use `collect(take(cycle(init), k))` here (as opposed to something simpler like filling the array)?

It seems like the reason to do this would be in case they gave us one number of initial conditions (2 for example) but wanted us to fill a matrix with a different number of columns (4 for example) -- In this case we would simulate 4 chains and start them with [first initial condition, second initial condition, first initial condition, second initial condition]. Is this what you had in mind? This should be elaborated on in the documentation, but I can imagine this surprising people.

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
+
+    ts_length, k = size(X)
+   
+    # if init is a vector, assign initial conditions to columns of X
+    # otherwise, just start each column at the same initial condition
+    real_init = Array(Int, k, )
+    if length(init) > 1
+        real_init = collect(take(cycle(init), k))
+    else
+        real_init = round(Int, init.*ones(k, ))
```
cc7768: Instead of `round(Int, init.*ones(k))` we could tell it that we want the vector to have type `Int` by doing `init .* ones(Int, k)`

cc7768:
Much thanks for pushing this through -- It looks good. We should check what is causing the failed (on a quick glance it looks like in the tests we use a `LinSpace` as the state space being output by a Markov approximation and somwhere the code is treating it as an `Array`)

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
-    init = rand(1:size(mc.p, 1), num_reps)
-    return simulate(mc, ts_length, init)
-end
 
-"""
-Fill `X` with sample paths of the Markov chain `mc` as columns.
-
-##### Arguments
+function simulate!(mc::MarkovChain, X::Matrix;
+                   init=rand(1:n_states(mc), size(X, 2)))
```
spencerlyon2: I think we should make `init` a positional argument instead of a keyword argument. The reason is that functions are not specialized on the type of keyword arguments. Normally this is fine because we can just specify the type of the argument, but here we want to allow for (at least) both `Integer` and `Vector{Integer}`. 

This, however, introduces some inconsistency with the API of `simulate` and `simulate!`.  I'd like some other thoughts on this before we make a final decision.

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
-function simulate(mc::MarkovChain, ts_length::Int; num_reps::Int=1)
-    init = rand(1:size(mc.p, 1), num_reps)
-    return simulate(mc, ts_length, init)
-end
 
-"""
-Fill `X` with sample paths of the Markov chain `mc` as columns.
-
-##### Arguments
+function simulate!(mc::MarkovChain, X::Matrix;
```
spencerlyon2: It is convention in Julia that the argument that is to be changed in a mutating function is the first argument of that function (e.g. see `rand!`, `copy!`, `fill!`, `A_mul_B!`, `A_mul_Bt!`, `A_mul_Bc!`, `At_mul_B!`, `Ac_mul_B!`, `At_mul_Bt!`, `Ac_mul_Bc!`, etc...)

I think we should do that here, so we'd move `X` to be the first argument. Also It makes sense to me to have the type be `X::Union{AbstractVector,AbstractMatrix}`

src/markov/mc_tools.jl
spencerlyon2: I agree with @cc7768 here. It doesn't actually make a difference in how the code runs, but having the trailing `, ` isn't consistent with the style in the rest of QuantEcon.jl

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
+    
+    # if init is a vector, assign initial conditions to columns of X
+    # otherwise, just start each column at the same initial condition
+    real_init = Array(Int, k, )
+    if length(init) > 1
+        real_init = collect(take(cycle(init), k))
+    else
+        real_init = round(Int, init.*ones(k, ))
+    end
+    X[1, :] = mc.state_values[real_init]'
```
spencerlyon2: No need for transpose here.

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
+- `X::Matrix{Int}` : Preallocated matrix to be filled with indices
+of the sample paths of the Markov chain `mc`.
+- `;init=rand(1:n_states(mc))` : Can be one of the following
+    - blank: random initial condition for each chain
+    - scalar: same initial condition for each chain
+    - vector: cycle through the elements, applying each as an
+      initial condition
+"""
+
+function simulate_indices!(mc::MarkovChain, X::Matrix{Int};
```
spencerlyon2: Same comment about having X come first. I also think we can loosen its type constraint so the `X` part of the function signature is

```Julia
simulate_indices!{T<:Integer}(X::Union{AbstractVector{T},AbstractMatrix{T}}, ...)
```

This will allow for all Integer subtypes and filling an vector-or-matrix-like thing.

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,155 @@ todense(::Type, A::Array) = A
...
+                           init=rand(1:n_states(mc), size(X, 2)))
```
spencerlyon2: Same note about `init` above

spencerlyon2:
Hi @vgregory757 thank you for working on this!

We've outlined some (fairly opinionated) changes that we'd like to see made as well as some items we'd like to discuss.

src/markov/mc_tools.jl
vgregory757: Yeah, that was the idea. I provided more clarification.

src/markov/mc_tools.jl
vgregory757: Couldn't we make `init` in `simulate` a positional argument over there as well?

vgregory757:
Thanks for the feedback @cc7768 and @spencerlyon2 -- I made the suggested changes

jstac:
@vgregory757 Thanks for making those changes.  With the latest I get the following error:

```
julia> P = [0.4 0.6; 0.2 0.8]
2x2 Array{Float64,2}:
 0.4  0.6
 0.2  0.8

julia> mc = MarkovChain(P, ["a", "b"])
Discrete Markov Chain
stochastic matrix of type Array{Float64,2}:
2x2 Array{Float64,2}:
 0.4  0.6
 0.2  0.8

julia> simulate(mc, 10)
ERROR: MethodError: `simulate!` has no method matching simulate!(::QuantEcon.MarkovChain{Float64,Array{Float64,2},Array{ASCIIString,1}}, ::Array{ASCIIString,2})
```
I could have this wrong but it seems that when we call `simulate!` [here](https://github.com/QuantEcon/QuantEcon.jl/pull/137/files#diff-16a555a93eb180b914e7f7be94d268f0R312) the order of the arguments is different from what's expected [here](https://github.com/QuantEcon/QuantEcon.jl/pull/137/files#diff-16a555a93eb180b914e7f7be94d268f0R335)

vgregory757:
@jstac It should work now. Sorry about that!

jstac:
@vgregory757 Thanks for that.  

Do you have time to look at that failing test by any chance?

spencerlyon2:
Hey @vgregory757 I believe that you are still getting test failures because you aren't exporting the proper functions. 

[This line](https://github.com/vgregory757/QuantEcon.jl/blob/8d905259c158dcd164cd3d999e57e4def46299c4/src/QuantEcon.jl#L47) should be changed so that it exports `simulate, simulate!, simulate_indices, simulate_indices!`

Thanks!

src/markov/mc_tools.jl
```diff
@@ -291,189 +291,156 @@ todense(::Type, A::Array) = A
...
+    if eltype(X) != eltype(mc.state_values)
+        throw(ArgumentError("Types in X must be the same as the types of the
+        Markov state values"))
+    end
 
-- `mc::MarkovChain` : MarkovChain instance.
-- `X::Matrix{Int}` : Preallocated matrix of integers to be filled with sample
-paths of the markov chain `mc`. The elements in `X[1, :]` will be used as the
-initial states.
+    ts_length, k = size(X)
```
spencerlyon2: This line will not actually work when x is an `AbstractVector`.

We should split it into

```julia
ts_length = size(X, 1)
k = size(X, 2)
```

jstac:
@vgregory757 Many thanks.  This is looking good.  I played around with it and everything seems fine.

One very small thing: in `simulate!`, the first argument is `X` and the second is `mc`.  I recommend the other way around, with `mc` first.  This matches the `simulate` function, where `mc` is the first argument.  It also matches the current documentation for `simulate!`.

Any objections?  If not would you mind to make that change?

spencerlyon2:
Hi @jstac, @vgregory757 originally had what you suggest here, but  I recommended the change to the current version.

See [this (hidden) comment](https://github.com/QuantEcon/QuantEcon.jl/pull/137#discussion_r75434961) above for my reasoning.

src/markov/mc_tools.jl
vgregory757: Done!

jstac:
@spencerlyon2 OK, thanks for letting me know.  That's fine then.  @vgregory757 Would you mind to swap the order of the arguments in the docstring of `simulate!` so that it matches? In particular [mc](https://github.com/QuantEcon/QuantEcon.jl/pull/137/files#diff-16a555a93eb180b914e7f7be94d268f0L347) should be below `X`.  Then I think we are done!

jstac:
@vgregory757 Many thanks.

All good on my end.

spencerlyon2:
Great work, thanks!

Issue #132, opened 2016-07-21
Changed meaning of .+= in Julia 0.5


stevengj:
Your code uses `x .+= y`, so you should know that in Julia 0.5 this has changed meaning to be equivalent to `broadcast!(identity, x, x .+ y)`, so that it mutates the `x` array (see JuliaLang/julia#17510 … in Julia 0.6 the whole operation will occur in-place without temporaries).  So `.+` should only be used if the left-hand side is a mutable array, and you don't mind mutating it.

At first glance, this seems like a problem for you, because you use `y .+= μ / (1 - ρ)` where `y` is a `LinSpace` object, which is immutable.  You can change it to `+=` to avoid trying to mutate `y`.   (However, the code is type-unstable in any case because `y` changes types, so you might want to use a new variable.)

spencerlyon2:
Thank you for opening this up! We will have to change what we are doing here.

PR #139, opened 2016-09-05
Update markov_approx.jl

colbec:
Tests fail in Julia 0.6 due to change in function of operator ".+="; so this proposal creates a new variable for the mean-variance standardized linspace vector and passes that on to the MarkovChain function.
Ref: https://github.com/QuantEcon/QuantEcon.jl/issues/132

### DIFF
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -73,13 +73,13 @@ function tauchen(N::Integer, ρ::Real, σ::Real, μ::Real=0.0, n_std::Integer=3)
     #       the cdf with a function that allows the distribution of input
     #       arguments to be [μ/(1 - ρ), 1] instead of [0, 1]
 
-    y .+= μ / (1 - ρ) # center process around its mean (wbar / (1 - rho))
+    yy = zeros(size(y)) .+ (μ / (1 - ρ)) # center process around its mean (wbar / (1 - rho)) in new variable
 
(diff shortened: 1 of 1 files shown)


src/markov/markov_approx.jl
```diff
@@ -73,13 +73,13 @@ function tauchen(N::Integer, ρ::Real, σ::Real, μ::Real=0.0, n_std::Integer=3)
     #       the cdf with a function that allows the distribution of input
     #       arguments to be [μ/(1 - ρ), 1] instead of [0, 1]
 
-    y .+= μ / (1 - ρ) # center process around its mean (wbar / (1 - rho))
+    yy = zeros(size(y)) .+ (μ / (1 - ρ)) # center process around its mean (wbar / (1 - rho)) in new variable
```
andreasnoack: Shouldn't this just be `yy = y .+ μ / (1 + ρ)`?

spencerlyon2: Yeah that makes more sense to me too.

@colbec would you be willing to make this change? Thanks

colbec: @spencerlyon2 Of course, thanks for the correction. I have changed my commit; I'm a bit of a beginner here with this process so please let me know if I have missed some part of the chain of events.

spencerlyon2:
Thank you for contributing @colbec. The PR is great!

PR #143, opened 2016-09-28
Minor issue with warnings

jstac:
If we look in [this notebook](http://nbviewer.jupyter.org/github/QuantEcon/QuantEcon.applications/blob/master/optgrowth/optgrowth_solutions_jl.ipynb) we see the warning

```
WARNING: max_iter exceeded in compute_fixed_point
```
It's a bit out of place because we deliberately set `max_iter`.  I can't think of a good fix so I just changed `exceeded` to `attained`.

### DIFF
--- a/src/compute_fp.jl
+++ b/src/compute_fp.jl
@@ -66,7 +66,7 @@ function compute_fixed_point{TV}(T::Function, v::TV; err_tol=1e-3,
     if iterate < max_iter && verbose
         println("Converged in $iterate steps")
     elseif iterate == max_iter
-        warn("max_iter exceeded in compute_fixed_point")
+        warn("max_iter attained in compute_fixed_point")
     end
 
     return v


spencerlyon2:
Sounds good to me will merge on green light from travis

oyamad:
A related issue:

Would it make sense to supply an option to set a *level* of verboseness? For example, `verbose = 0` suppresses all the warnings, `verbose = 1` shows some of the warnings, and `verbose = 2` shows all, etc. Currently, there is no way to make the max_iter warning not show up, but there are cases where one wants to iterate exactly k times, where k is a parameter of the algorithm (think of modified policy iteration).

spencerlyon2:
That makes sense to me. I'd be happy to accept a PR that adds that option to `compute_fixed_point`

jstac:
Good idea.  See PR #144

PR #146, opened 2016-11-02
ENH: added iterators for MarkovChain simulation

spencerlyon2:
This PR revamps the internals behind `simulate` and `simulate_indices` to use the Julia iterator protocol.

The current API for `simulate`, `simulate!`, `simulate_indices`, `simulate_indices!` has not changed. 

However, we can now do things more flexibly.  I have defined two new types `MCIndSimulator` and `MCSimulator` that are responsible for _iterating_ over the simulation of a MarkovChain. This enables us to do things like

```julia
mcs = MCSimulator(mc, 200, 1)

# get multiple simulations, using same initial condition
sim1 = collect(mcs)
sim2 = collect(mcs)

# iterate over a simulation path, without creating array to hold path
for i in mcs
    do_something_cool(i)
end
```

You can see the new (and greatly simplified!!) implementation of `simulate!` and `simulate_indices!` for other ideas about what is possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: julia
 sudo: false
 julia:
-    - 0.4
     - 0.5
     - nightly
 matrix:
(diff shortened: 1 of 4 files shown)


spencerlyon2:
The error here is because some parts of the [iterator interface ](http://docs.julialang.org/en/release-0.5/manual/interfaces/) is not defined for Julia 0.4 -- meaning some functions I implemented don't exist in Julia 0.4.

Is it time that we freeze QuantEcon on 0.4 and have new features going forward be for 0.5 only?

If so, I think we can adjust the travis script and then merge this

cc7768:
This looks really slick. I think it is a big improvement -- I'm in favor of freezing (discontinuing support) for Julia v0.4 and only doing 0.5 <- In my mind this is similar to our treatment of Python 2.

I will have a closer look at code over next day or so and provide some comments.

PR #151, opened 2016-11-14
add back 0..4 testing

tkelman:
if you don't support 0.4 any more, you need to update REQUIRE accordingly

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
 language: julia
 sudo: false
 julia:
+    - 0.4
     - 0.5
     - nightly
 matrix:
   allow_failures:
     - julia: nightly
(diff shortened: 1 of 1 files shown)


tkelman:
If this isn't going to work, then you need to modify the julia requirement of the tag on metadata, otherwise the package will break for all users on 0.4 if they Pkg.update.

spencerlyon2:
Tests fail on 0.4 only because type inference isn't able to infer the return value of one of my functions. The code itself runs -- it will just introduce type instabilities for 0.4 and not on 0.5+.

tkelman:
in that case, should only use inferred on 0.5.

spencerlyon2:
Done

PR #153, opened 2017-02-04
ENH: add linear interpolation routine

### LINKED ISSUE #152: 1D linear interpolation convenience function
I would like to have a convenience function that does 1D linear interpolation.

For example,

```
using QuantEcon
f = interp(grid, function_vals)
```

gives me a function `f` that uses linear interpolation between grid points.  To the left of the first grid point iit returns `function_vals[1]` and to the right of the largest grid point it returns `function_vals[end]`.

I know that `Interpolations` can be used but the syntax is hard to remember and the returned object is a bit idiosyncratic.  

The function can be hand coded and still give optimal performance.

sglyon:
Closes #152

### DIFF
--- a/src/QuantEcon.jl
+++ b/src/QuantEcon.jl
@@ -9,7 +9,7 @@ using Distributions
 import Distributions: pdf, skewness, BetaBinomial
 using DSP: TFFilter, freqz
 using Primes: primes
-using Compat
+using Compat: view, @compat
 
 # useful types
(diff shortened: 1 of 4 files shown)


jstac:
It works nicely in my use case.

https://gist.github.com/anonymous/5c7f75fda240bc03ca392cfccffec4be

I think it's ready to merge.  Thanks again.

sglyon:
Excellent! That does make things simpler.

PR #154, opened 2017-02-07
Sl/0.6

sglyon:
This gets QuantEcon.jl ready for Julia 0.6.

It mostly includes cosmetic changes to use broadcasting syntax (`f.(...)`) instead of vectorized functions (`f(x...)`) and change `Array(T, ...)` to `Array{T}(...)`.

There are three substantial changes that might require changes to lectures:

1. We no longer use task based iteration for the moment sequence of the LSS type. Instead we use a new (non exported) type `LSSMoments` and define the `start`, `next`, and `done` methods for it. There is one commit dedicated to this change, so if you want to see how that changes other code take a look at that.
2. I've deprecated the `ecdf(::ECDF, x)` function in favor of a `call` method on the type. So  now instead of writing `ecdf(e, x)` we write `e(x)`. This is a true deprecation so the old syntax will still work (for now), but will emit a warning
3. We've officially dropped support for Julia 0.4. With 0.6 around the corner and 1.0 due this summer I think supporting 0.4 is not necessary. If anyone has objections to this we can still support it if we really want to, but will require us to sprinkle `@compat` in many places across the lib

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,6 @@ sudo: false
 julia:
     - 0.5
     - nightly
-matrix:
-  allow_failures:
-    - julia: nightly
 notifications:
     email: false
 #script: # use the default script setting which is equivalent to the following
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: julia
 sudo: false
 julia:
-    - 0.4
     - 0.5
     - nightly
 matrix:
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.4
+julia 0.5
 DSP
 Distributions 0.8.9
 LightGraphs
--- a/REQUIRE
+++ b/REQUIRE
@@ -3,5 +3,5 @@ DSP
 Distributions 0.8.9
 LightGraphs
 Primes
-Compat 0.8.4
+Compat 0.18.0
 StatsBase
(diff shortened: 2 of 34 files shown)


sglyon:
cc @jstac  for potential changes to lectures

jstac:
Thanks for the heads up.  I've run some `grep` searches and there are no problems vis-a-vis the lectures.  

I prefer the new syntax for initializing arrays and I like your change to `ECDF`.

All good from my end.

sbromberger:
Hi all,

LightGraphs is preparing for a fairly large upgrade with Julia 0.6 as well, and I wanted to make you aware of what we're doing. Very few things are breaking, but there will be some potential performance improvements that you might be able to take advantage of.

In a nutshell: we're abstracting `Graph` and `DiGraph`, renaming them to `SimpleGraph` and `SimpleDiGraph`*, and parameterizing vertex indices to `<: Integer`. This will allow you to create, e.g.,  `SimpleGraph{UInt8}`s that will be much more space-efficient if you have graphs that are < 256 vertices.

* While we plan on more abstractions (e.g., for weighted graphs) in the future, for now `Graph` and `DiGraph` will continue to work and will continue to default to `SimpleGraph`, and the default parameterization for both `Simple(Di)Graph`s and `(Di)Graph`s will be `Int`.

Feel free to review the (very long) WIP at https://github.com/JuliaGraphs/LightGraphs.jl/pull/541 - your feedback would be greatly appreciated.

PR #156, opened 2017-03-31
ENH: Change markov approximation means

cc7768:
We currently represent Markov chains that approximate the following equation

y_{t+1} = \mu + \rho y_{t} + \sigma \varepsilon_{t+1}

I think it is more intuitive to approximate 

y_{t+1} = (1-\rho) \mu + \rho y_{t} + \sigma \varepsilon_{t+1}

The reason I think this is more intuitive is because the mean of the first process is (1-\rho) \mu which doesn't match with our documentation which says that the argument mu is the mean of the AR(1) process. If we change the process we are representing to the one I propose then mu is actually the mean of the AR(1) process. (Let me know if that description is a bit confusing -- I was a bit repetitive in the language I used)

What do other people think?

### DIFF
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -88,7 +88,7 @@ Rouwenhorst's method to approximate AR(1) processes.
 
 The process follows
 
-    y_t = μ + ρ y_{t-1} + ε_t,
+    y_t = (1-ρ)μ + ρ y_{t-1} + ε_t,
 
 where ε_t ~ N (0, σ^2)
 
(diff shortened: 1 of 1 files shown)


sglyon:
Will you add one test that:

- uses tauchen to define a Markov chain with mu != 0
- compute the stationary distribution of the Markov chain. Call it, for example, `sd`.
- ensure that `dot(sd, mc.state_values)` is equal to mu

?

jstac:
@cc7768 I'm in favor of staying with the current form, which is a bit simpler and more common.  We could just call \mu the *drift* instead of the mean.

cc7768:
Will add the test later today.

cc7768:
Calling it the drift is an improvement I can live with. I'll make that change (and add a version of the test @sglyon) later today -- Unless someone beats me to it.

I agree the current form is probably a bit simpler -- I just really like that the proposed form makes it so that you are just inputting the mean you'd like for your process.

PR #157, opened 2017-04-18
Adding MVNSampler and method to draw for singular covariance matrix

Shunsuke-Hori:
@jstac @sglyon This PR adds `MVNSampler`, which contains mean vector `mu`, covariance matrix `Sigma`, and `Q` which is the square root of `Sigma`. `Q` is computed with eigen decomposition

Method to draw by `rand` is also added

Finally, `LSS` type can have `MVNSampler` as `dist` only when covariance matrix is positive semi-definite


jstac:
@Shunsuke-Hori This looks nice.  Thanks.

I see you've added and removed a test.  It would be good to have one.  Can you please let us know when this is ready for review?

Shunsuke-Hori:
@jstac Sure. 

Something was wrong in my test file, so I removed it. I will retry it soon.

Shunsuke-Hori:
@jstac @sglyon Do you have any good idea to quickly test whether `rand` draws from the true distribution?

Relying on LLN, I confirmed that the mean of samples are approaching to `mu` but it is very time-consuming to draw enough samples

For now, the PR only tests the type of the draw

Shunsuke-Hori:
Once the above issue regarding the test is solved, the PR is ready for review

sglyon:
I think it would be sufficient to ensure that the matrix `Q` is in fact the matrix square root of the the given `Sigma` matrix. 

I'm not sure that the algorithm implemented here does this.

See this example:

```
julia> Sigma_0 = [0.0  0.0      0.0      0.0;
                  0.0  5.26316  4.73684  0.0;
                  0.0  4.73684  5.26316  0.0;
                  0.0  0.0      0.0      0.0];

julia> _Q1 = sqrtm(Sigma_0)  # algorithm uses complex Shcur decomposition
4×4 Array{Complex{Float64},2}:
 0.0+0.0im      0.0+0.0im      0.0+0.0im  0.0+0.0im
 0.0+0.0im  1.94388+0.0im   1.2184+0.0im  0.0+0.0im
 0.0+0.0im   1.2184+0.0im  1.94388+0.0im  0.0+0.0im
 0.0+0.0im      0.0+0.0im      0.0+0.0im  0.0+0.0im

julia> all(isreal, _Q1)  # check that all are real
true

julia> Q1 = real(_Q1);  # check passed -- just keep the real part

julia> Q2 = begin  # now use algorithm in this PR
       lambdas,U = eig(Sigma_0)          # eigen decomposition
       LAMBDAr = diagm(broadcast(sqrt,lambdas))
       U * LAMBDAr
       end;

julia> Q1
4×4 Array{Float64,2}:
 0.0  0.0      0.0      0.0
 0.0  1.94388  1.2184   0.0
 0.0  1.2184   1.94388  0.0
 0.0  0.0      0.0      0.0

julia> Q2
4×4 Array{Float64,2}:
 0.0  0.0   0.0       0.0
 0.0  0.0   0.512991  2.23607
 0.0  0.0  -0.512991  2.23607
 0.0  0.0   0.0       0.0
```

What do people think about this discrepancy. It could be  just an issue with non-uniqueness of eigenvectors, but I'm a bit concerned that the non-zero elements show up in the `[2:3, 3:4]` block of `Q2` instead of the `[2:3, 2:3]` block like they do in `Q1` and `Sigma_0`.

jstac:
@sglyon Good catch.  I'm not sure why off the top of my head.

@Shunsuke-Hori Would it be better to use `sqrtm`?  I think the code would be clearer.

cc7768:
FWIW

```
julia> Q1*Q1'
4×4 Array{Float64,2}:
 0.0  0.0      0.0      0.0
 0.0  5.26316  4.73684  0.0
 0.0  4.73684  5.26316  0.0
 0.0  0.0      0.0      0.0

julia> Q2*Q2'
4×4 Array{Float64,2}:
 0.0  0.0      0.0      0.0
 0.0  5.26316  4.73684  0.0
 0.0  4.73684  5.26316  0.0
 0.0  0.0      0.0      0.0
```

jstac:
@cc7768 Thanks for clarifying.  

`sqrtm` seems best in terms of readability.

Shunsuke-Hori:
@sglyon @cc7768 @jstac Thank you everyone. All comments are really helpful for me.

Now, the PR uses Schur decomposition

Since the test file only checks `Sigma == Q*Q'`, `draw` from `MVNSampler` is not covered, which reduces coverage

oyamad:
I haven't followed the background behind what this PR is going to add, but let me ask a few questions:

* What is the assumption about the input `Sigma_0`? Is it that it is assumed to be (symmetric and) positive semi-definite already, and `elseif any(eig(Sigma_0)[1].==0.0)` checks whether it is singular?

* (Depending on the algorithm used internally) is it possible that `eig(Sigma_0)` does some kind of decomposition already so that doing `sqrtm` again is redundant to some extent?

* What happens if a singular positive-semi definite covariance matrix is passed to `MultivariateNormal`? Would `try MultivariateNormal(mu_0, Sigma_0) catch ...` work?

Shunsuke-Hori:
@oyamad The lecture "Permanent income II" requires a draw from multivariate normal with **singular** covariance matrix.

However, `MultivariateNormal` from `Distributions.jl` only allows **positive definite** covariance, which is not necessary for sampling. So, I am trying to add a method for that.

Regarding your question, 
* `Sigma_0` is assumed to be positive semi-definite (= the requirement for covariance matrix). 
If any of the eigenvalue is zero, i.e. if it is not positive definite, use `MVNSampler` instead of `MultivariateNormal` (maybe I should fix the comment in the code).
* We can also obtain square root of `Sigma_0` by `eig` and some matrix operations. (`Q2` in [this comment](https://github.com/QuantEcon/QuantEcon.jl/pull/157#issuecomment-295292723)) It would be better if I could reduce the use of `eig` and `sqrtm` keeping the code easy to read. But I have no idea how to write such code.
* `MultivariateNormal(mu_0, Sigma_0)` does not works if `Sigma_0` is singular

Please let me know if I misunderstand something and if you have any idea for improvements.

oyamad:
@Shunsuke-Hori Thanks.

Would the following work then?

```jl
try
    dist = MultivariateNormal(mu_0, Sigma_0)
catch PosDefException
    dist = MVNSampler(mu_0, Sigma_0)
end
```

(Some name like `SingularMultivariateNormal`/`SingularMvNormal` or `DegenerateMultivariateNormal`/`DegenerateMvNormal` might be clearer than `MVNSampler`.)

sglyon:
I'm going to take a closer look in the morning, but do we need to use any features that `MultivariateNormal` provides beyond sampling? 

If not, a viable option could be to just always construct/use a `MVNSampler` inside an instance of `LSS`. Then we don't have to worry about these issues.

If we do go that route we will definitely want to make sure that the assumptions we are implicitly making (e.g. symmetry and psd) are enforced in the `MVNSampler` constructor

Shunsuke-Hori:
@oyamad Thank you for the suggestion and sorry to miss it before. Since `try-catch` introduces scope, the following snippet works. (is the first line, `dist=nothing`, a good way to go?)

    dist=nothing
    try
        dist = MultivariateNormal(mu_0, Sigma_0)
    catch PosDefException
        dist = MVNSampler(mu_0,Sigma_0)
    end

On the other hand, I agree with @sglyon. It seems that the sampling is the only feature we need and it would be possible to always use `MVNSampler` 

**In short,**
1. Which is better?
    1. use `try-catch` and have other information about distribution if possible
    2. always use `MVNSampler` 
1. What is the best name for the new type if we choose 1.ii.?

oyamad:
Always using `MVNSampler` seems strictly better.

Then one way to verify the assumption and construct an instance is:
* check that the input matrix is symmetric, and
* do eigenvalue decomposition to check that all the eigenvalues are nonnegative (possibly allowing some tolerance) and get the square root matrix.

oyamad:
Or maybe you can directly use `cholfact(A, [uplo::Symbol, ]Val{true}; tol = 0.0)`, next to [this](https://docs.julialang.org/en/stable/stdlib/linalg/#Base.cholfact).

sglyon:
I would propose something like this:

```julia
type MVNSampler{T<:Real}
    mu::Array{T}
    Sigma::Array{T}
    Q::Array{T}     # square root of Sigma 
end

function MVNSampler(mu::AbstractVector,Sigma::AbstractArray)
    try
        cholfact(Sigma, :U, Val{true})
    catch
        error("Sigma must be symmetric and positive semi-definite")
    end
    _Q = sqrtm(Sigma)  # complex Schur decomposition
    
    !all(isreal, _Q) && error("TODO: write down what this means!!!")

    Q =real(_Q)        # square root of Sigma
    return MVNSampler(mu,Sigma,Q)
end
```

and then always only using `MVNSampler` inside LSS

oyamad:
@sglyon 
* `cholfact(Sigma, :U, Val{true})` does not throw an error even if `Sigma` is (symmetric and) non positive semidefinite. Try for example `cholfact([0 -1; -1 0], :U, Val{true})`.
* I don't think it necessary to do Factorization more than once.

Here's my proposal:

```jl
immutable MVNSampler{TM<:Real,TS<:Real}
    mu::Vector{TM}
    Sigma::Matrix{TS}
    Q::Matrix{Float64}
end

const TOL1 = 1e-10  # some small value
const TOL2 = 1e-10  # some small value

function MVNSampler{TM<:Real,TS<:Real}(mu::Vector{TM}, Sigma::Matrix{TS})
    n = length(mu)
    
    # TODO: check Sigma is an n x n symmetric matrix
    
    C = cholfact(Symmetric(Sigma, :L), Val{true})
    
    if C.rank == n  # Positive definite
        Q = tril(C.factors)[C.piv, C.piv]
        return MVNSampler(mu, Sigma, Q)
    end
    
    const non_PSD_msg = "Sigma must be positive semidefinite"
    
    for i in C.rank+1:n
        if C[:L][i, i] < -TOL1  # Not positive semidefinite
            throw(ArgumentError(non_PSD_msg))
        end
    end
    
    Q = tril(C.factors)[C.piv, C.piv]
    if maxabs(Sigma - Q * Q') > TOL2  # Not positive semidefinite
        throw(ArgumentError(non_PSD_msg))
    end
    return MVNSampler(mu, Sigma, Q)
end
```

oyamad:
`MVNSampler` will be helpful itself.

Feature requests for `rand(::MVNSampler)`:

* Add an `rng` option.
* Add a `num_samples` option, which will return an `n` x `num_samples` matrix.

Thanks!

sglyon:
Thank you very much @oyamad!

That example looks great to me.

I have a couple comments:

- We can move the two `TOL` variables into the constructor so they aren't global variables 
- no variables in the function need to be declared `const`, so we can remove that from the `TOL` variables as well as `non_PSD_msg`.
- Do we know that the element type of `Q` must be `Float64`? What happens when `TS` is, for example, `Float32`?

I think that we can emulate the interface of `rand` and define the following methods:

```julia
rand(::MVNSampler) --> single sample as a `Vector`
rand(::MVNSampler, ns::Integer...) --> Many samples as a Array{T,length(ns)+1}

# and methods with the optional rng argument first
rand(rng::AbstractRNG, ::MVNSampler) --> single sample as a `Vector`, using png
rand(rng::AbstractRNG, ::MVNSampler, n::Integer...) --> Many samples as a Array{T,length(n)+1}
```

oyamad:
@sglyon Good points, thanks.

It will be `Float32` if `TS` is `Float32` according to the code of `pstrf!`.
So the type definition should be:

```jl
immutable MVNSampler{TM<:Real,TS<:Real,TQ<:LinAlg.BlasReal}
    mu::Vector{TM}
    Sigma::Matrix{TS}
    Q::Matrix{TQ}
end
```

Regarding the values of `TOL`s, I have no idea what the best values are... (In principle, they should be a relative tolerance.)

oyamad:
I think `MVNSampler` deserves an independent `.jl` file (not as part of `lss.jl`). I would like to use it in [`Games.covariance_game`](https://github.com/QuantEcon/Games.jl/pull/24).

Shunsuke-Hori:
Thank you @sglyon and @oyamad. I will integrate your codes and ideas.

How about adding `MVNSampler` into `util.jl`?

sglyon:
For now my preference would be to just out it in its own file. 

Thank you for working on this!

Shunsuke-Hori:
Now, `MVNSampler` is in `sampler.jl` and `LSS` always uses it.

- For the symmetry checking of Sigma, julia v0.4 and v0.5 have different function names. It is dealt by the code starting from [this line](https://github.com/QuantEcon/QuantEcon.jl/pull/157/files#diff-e5a4b188a845b3ee7871e0470784c745R23)
- `sortperm` is added in the line to construct `Q`  ([here](https://github.com/QuantEcon/QuantEcon.jl/pull/157/files#diff-e5a4b188a845b3ee7871e0470784c745R37) and [here](https://github.com/QuantEcon/QuantEcon.jl/pull/157/files#diff-e5a4b188a845b3ee7871e0470784c745R49))
- The [code of `rand` for many sampling](https://github.com/QuantEcon/QuantEcon.jl/pull/157/files#diff-e5a4b188a845b3ee7871e0470784c745R60) is not beautiful. The procedure is
    - transform `ns`, which is `Tuple`, into `Array`
    - add one dimension of `length(d.mu)`
    - transform `Array` into `Tuple`
    - draw by `randn`  (the draw can be multidimensional array)
    - reshape the draw into matrix to make multiplication possible
    - reshape the result of multiplication into appropriate dimensions

oyamad:
Sorry, `Q = tril(C.factors)[C.piv, C.piv]` should have been

```jl
p = invperm(C.piv)
Q = tril(C.factors)[p, p]
```

src/sampler.jl
```diff
@@ -0,0 +1,67 @@
...
+    Sigma_is_symmetric = nothing
+    try
+        Sigma_is_symmetric = issymmetric(Sigma) # v0.5
+    catch
+        Sigma_is_symmetric = issym(Sigma)       # v0.4
+    end
+
+    if Sigma_is_symmetric == false # Check Sigma is symmetric
+        throw(ArgumentError("Sigma must be symmetric"))
+    end
```
oyamad: You can use Compat.jl:

```jl
@compat issymmetric(Sigma) || throw(ArgumentError("Sigma must be symmetric"))
```

But do we still want to keep 0.4 support? @sglyon

src/sampler.jl
```diff
@@ -0,0 +1,67 @@
...
+
+immutable MVNSampler{TM<:Real,TS<:Real,TQ<:LinAlg.BlasReal}
+    mu::Vector{TM}
+    Sigma::Matrix{TS}
+    Q::Matrix{TQ}
+end
+
+function MVNSampler{TM<:Real,TS<:Real}(mu::Vector{TM}, Sigma::Matrix{TS})
+    TOL1 = 1e-10  # some small value
+    TOL2 = 1e-10  # some small value
```
oyamad: These values should be adjusted to "optimal" values.

Shunsuke-Hori: Sorry, I have no idea what "optimal" means. 

Does it mean the tolerance should be relative value?

oyamad: Do numerical experiments by generating singular positive semidefinite matrices and their "slightly non positive semidefinite" perturbations to check the diagonal values of the outputs of their cholfacts and the difference `maxabs(Sigma - Q * Q')`.

Shunsuke-Hori:
Thank you @oyamad for your help. 

Regarding the tolerance value, diagonal element of `cholfact` never returns negative value.

On the other hand, `maxabs(Sigma - Q * Q')` can be different from zero and the maximum difference is `6.3e-13`, so I set `1e-12` to `TOL2`

oyamad:
@Shunsuke-Hori 
* What set of matrices did you try (what sizes, how you generated them, etc.)?
* Is the difference `maxabs(Sigma - Q * Q')` more or less constant if you multiply `Sigma`, say by 10?

Shunsuke-Hori:
As `Sigma`, I constructed 100x100 matrix whose `[51:end, 51:end]` elements are square of `randn((50,50))` and the other elements are zero

I drew samples of such `Sigma` `1e4` times

The difference changes depending on the scale of `Sigma`, but it never becomes larger than 1e-12

    N =1e4
    min_diag,max_dif=100,0
    Sigma=zeros(Float64,100,100)
    for i =1:N
        randmat=randn((50,50))
        Sigma[51:end,51:end].= randmat * randmat'
        all(eig(Sigma)[1] .>=0.0) || error("Sigma is not positive semi definite")
        C= cholfact(Symmetric(Sigma, :L), Val{true})
        p = invperm(C.piv)
        Q = tril(C.factors)[p,p]
        diag_element = diag(C[:L])
        dif = maxabs(Sigma - Q * Q')
        min_diag = ifelse(min_diag>minimum(diag_element),minimum(diag_element),min_diag)
        max_dif = ifelse(max_dif<maximum(dif),maximum(dif),max_dif)
    end
    
    println("min_diag = $min_diag, max_dif = $max_dif")

oyamad:
It is fine, if that's the only usecase in which a singular psd matrix is used. (In this case, it is obvious, according to the algorithm, that the diagonals are always nonnegative (almost with probability one).)

You might play with this:

```jl
function diag_min(rng, n; num_reps=1, a=1)
    out = Array(Float64, (num_reps, 2))
    for i in 1:num_reps
        A = randn(rng, n, n)
        A[:, end] = sum(A[:, 1:end-1], 2)
        Sigma = A * A' * a
        C = cholfact(Symmetric(Sigma, :L), Val{true})
        out[i, 1] = minimum(diag(C.factors))  # Abolute value
        out[i, 2] = out[i, 1] / C.factors[1, 1]  # Relative value
    end
    return out
end
```

As an example:

```jl
rng = MersenneTwister(0)
diag_min(rng, 100; num_reps=10, a=100)
```

```jl
10×2 Array{Float64,2}:
 -8.18545e-12  -2.56589e-14
  2.52319e-6    7.65067e-9 
 -3.72893e-11  -1.39294e-13
 -1.33241e-10  -4.71214e-13
 -5.91172e-12  -1.58064e-14
 -8.18545e-12  -3.25311e-14
  4.54874e-6    1.78092e-8 
 -5.77529e-11  -1.52844e-13
 -7.93534e-11  -3.32885e-13
  7.44843e-6    3.41831e-8
```

Here, the rank is reduced by a specific manner, and only by 1.

Shunsuke-Hori:
@oyamad I did some experiments with your matrices and found that `TOL1` and `TOL2`  should be `1e-5` and `1e-10`, respectively.

oyamad:
`TOL1 = 1e-5` sounds too big to me, but you should have a justification.

Shunsuke-Hori:
@oyamad Sorry, I was looking opposite sign. 

But still, the `TOL1` would be `1e-8` because minimum of `out[i, 1]` takes smaller than `-1e-9`

oyamad:
What about the relative value?

Shunsuke-Hori:
minimum relative value takes smaller than `1e-12`

src/sampler.jl
sglyon: Hey guys, let's drop 0.4 support.

No need for this

Shunsuke-Hori:
`@compat` is removed

sglyon:
Hey @Shunsuke-Hori thank you for all the work on this.  Have we converged on "correct" values for the tolerance levels? 

I have one more request to make. I think I bit off a bit more than we need to chew when I requested the uber-general `rand` methods. I think all we need to supply is what @oyamad originally suggested:

```julia
rand(::MVNSampler)  -> Vector
rand(::MVNSampler, n::Integer)  -> Matrix
```

I checked and this is the same interface provided by the multivariate distributions in Distributions.jl.

@Shunsuke-Hori are you willing to make this change? 


Once we merge this PR and #154 I want to make a release of QuantEcon.

Shunsuke-Hori:
I took a look at [`numpy` code](https://github.com/numpy/numpy/blob/v1.12.0/numpy/random/mtrand/mtrand.pyx#L4358) and realized that absolute and relative tolerance are defined in [this way](https://github.com/numpy/numpy/blob/v1.12.0/numpy/core/numeric.py#L2543) (Thank you @oyamad for this advice)

However, I don't know how to set `atol` and `rtol` at the same time (and why `numpy` set 1e-8)

Instead, I did some experiments and estimated `atol` and `rtol` separately assuming that `rtol` and `atol` is zero respectively

I found that the relative value of the difference `Sigma-Q*Q'` is very robust to the scale change of `Sigma` (`RTOL2` in this PR) while other tolerance are not so robust as `RTOL2`

So now, this PR uses relatively loose criteria for `ATOL1`, `RTOL1`, and `ATOL2` and relatively strict  criterion for `RTOL2` and requires to satisfy `RTOL2`

src/sampler.jl
```diff
@@ -0,0 +1,60 @@
...
+    if C.rank == n  # Positive definite
+        p = invperm(C.piv)
+        Q = tril(C.factors)[p,p]
+        return MVNSampler(mu, Sigma, Q)
+    end
+
+    non_PSD_msg = "Sigma must be positive semidefinite"
+
+    for i in C.rank+1:n
+        if C[:L][i, i] < -ATOL1 || C[:L][i, i]/norm(diag(C[:L])) < -RTOL1 # Not positive semidefinite
```
oyamad: Don't do division.

Try this:

```jl
        if C[:L][i, i] < -ATOL1 - RTOL1 * C[:L][1, 1]
```

(I have no logical justification, but propose to make it relative just to the largest diagonal element.)

You will save one line if you do this:

```jl
    for i in C.rank+1:n
        C[:L][i, i] >= -ATOL1 - RTOL1 * C[:L][1, 1] ||
            throw(ArgumentError(non_PSD_msg))
    end
```

src/sampler.jl
```diff
@@ -0,0 +1,60 @@
...
+    return MVNSampler(mu, Sigma, Q)
+end
+
+# methods to draw from `MVNSampler`
+Base.rand{TM,TS,TQ}(d::MVNSampler{TM,TS,TQ}) = d.mu + d.Q * randn(size(d.mu))
+Base.rand{TM,TS,TQ}(d::MVNSampler{TM,TS,TQ}, n::Integer) = d.mu.+d.Q*randn(length(d.mu),n)
+
+# methods with the optional rng argument first
+Base.rand{TM,TS,TQ}(rng::AbstractRNG, d::MVNSampler{TM,TS,TQ}) = d.mu + d.Q * randn(rng, size(d.mu))
+Base.rand{TM,TS,TQ}(rng::AbstractRNG, d::MVNSampler{TM,TS,TQ}, n::Integer...) = d.mu.+d.Q*randn(rng,(length(d.mu),n))
```
sglyon: Will you please delete the `...` after `n::Integer`?

src/sampler.jl
```diff
@@ -0,0 +1,60 @@
...
+    for i in C.rank+1:n
+        if C[:L][i, i] < -ATOL1 || C[:L][i, i]/norm(diag(C[:L])) < -RTOL1 # Not positive semidefinite
+            throw(ArgumentError(non_PSD_msg))
+        end
+    end
+
+    p = invperm(C.piv)
+    Q = tril(C.factors)[p,p]
+
+    if maxabs(Sigma - Q * Q') > ATOL2 && maxabs(Sigma - Q * Q')/max(norm(Sigma),norm(Q * Q')) > RTOL2 # Not positive semidefinite
```
oyamad: 1. If we want to exactly follow the [NumPy code on `master`](https://github.com/numpy/numpy/blob/master/numpy/random/mtrand/mtrand.pyx#L4506) (which is more recent than the code in the link provided by @Shunsuke-Hori), it will be:
   
   ```jl
   Q_2 = Q * Q'
   for j in 1:n
       for i in 1:n
           abs(Q_2[i, j] - Sigma[i, j]) <= ATOL2 + RTOL2 * abs(Sigma[i, j]) ||
               throw(ArgumentError(non_PSD_msg))
       end
   end
   ```

   where `ATOL2` and `RTOL2` default to `1e-8`.

2. Alternatively, we can just use [`isapprox`](https://docs.julialang.org/en/stable/stdlib/math/#Base.isapprox):
   
   ```jl
   isapprox(Q*Q', Sigma; rtol=RTOL2, atol=ATOL2) ||
       throw(ArgumentError(non_PSD_msg))
   ```

src/sampler.jl
```diff
@@ -0,0 +1,60 @@
...
+        throw(ArgumentError(non_PSD_msg))
+    end
+
+    return MVNSampler(mu, Sigma, Q)
+end
+
+# methods to draw from `MVNSampler`
+Base.rand{TM,TS,TQ}(d::MVNSampler{TM,TS,TQ}) = d.mu + d.Q * randn(size(d.mu))
```
oyamad: I would use `randn(length(d.mu))` instead of `randn(size(d.mu))`, to be consistent with the line below.

src/sampler.jl
```diff
@@ -0,0 +1,60 @@
...
+Base.rand{TM,TS,TQ}(d::MVNSampler{TM,TS,TQ}, n::Integer) = d.mu.+d.Q*randn(length(d.mu),n)
```
oyamad: You might define `rand(d)` and `rand(d, n)` by

```jl
Base.rand(d::MVNSampler) = rand(Base.GLOBAL_RNG, d)
Base.rand(d::MVNSampler, n::Integer) = rand(Base.GLOBAL_RNG, d, n)
```

after defining `rand(rng, d)` and `rand(rng, d, n)`.

Do we need type parameters `{TM,TS,TQ}`?

Shunsuke-Hori: > Do we need type parameters `{TM,TS,TQ}`?
 I do not think so. I will  remove it.

oyamad:
You also need a test file `test_sampler.jl`:

* With a positive definite `Sigma`, check that (it is accepted and) `Q*Q'` is close to `Sigma`.
* With a singular positive semi-definite `Sigma`, check that (it is accepted and) `Q*Q'` is close to `Sigma`.
  You might try:
  - `zeros(n, n)`, for which you can test that the outcome of `rand` is equal to `mu`
  - `ones(n, n)`, for which you can test that the outcome of `rand` is a constant vector
  - matrix with `1` on the diagonal and `-1/(n-1)` otherwise, for which you can test that the sum of the components in the outcome is equal to `0`
* Check that a non positive semi-definite `Sigma` is rejected.

Shunsuke-Hori:
@sglyon Passing zero matrix for `Sigma` is fine. The mean is always returned in this case.

When it is ready to re-review, I will comment again.

Shunsuke-Hori:
When I try `ones(n,n)` as `Sigma` for n>2, `Sigma` is different from  `Q * Q'` (not because of computational error)

But I could not figure out what is wrong. Could you give me some advice?

oyamad:
@Shunsuke-Hori Sorry, I was misunderstanding the algorithm.
Will you try this?

```jl
    A = copy(Sigma)
    C = cholfact!(Symmetric(A, :L), Val{true})
    r = C.rank
    p = C.piv

    if r == n  # Positive definite
        Q = tril!(A)[p, p]
        return MVNSampler(mu, Sigma, Q)
    end

    ...

    tril!(view(A, :, 1:r))
    A[:, r+1:end] = 0
    Q = A[p, p]
```

And will you check if this works for random matrices of rank 1, ..., n-2?

Shunsuke-Hori:
@oyamad Thank you for the help, but it seems something is still wrong

`Q * Q'` is different from `Sigma`

oyamad:
Sorry, of course `p = C.piv` should have been `p = invperm(C.piv)`.

Shunsuke-Hori:
Thank you very much @oyamad, I should have noticed that

Now, it is ready for review

sglyon:
@Shunsuke-Hori and @oyamad thank you for all the work on this.

@Shunsuke-Hori will you please remove Julia 0.4 support. To do this we need to do 2 things:

1. Make [this line](https://github.com/QuantEcon/QuantEcon.jl/blob/master/REQUIRE#L1) read `Julia 0.5`
2. Remove [this line](https://github.com/QuantEcon/QuantEcon.jl/blob/855dcbfb02db11f5b0efcd864e32b016c8865aa6/.travis.yml#L4)


Then we can make sure tests are passing before merging this.

Thanks

Shunsuke-Hori:
@sglyon Done.

sglyon:
Looks good to me.

@oyamad when you are happy with this please feel free to merge.

oyamad:
@Shunsuke-Hori Thanks.

* [ ] `test_sampler.jl` is actually not tested. You have to add it to [runtests.jl](https://github.com/QuantEcon/QuantEcon.jl/blob/master/test/runtests.jl).
* [ ] In passing, will you replace [these lines](https://github.com/QuantEcon/QuantEcon.jl/blob/master/test/runtests.jl#L5-L10) with just `using Base.Test`?
* [ ] Also change [this line](https://github.com/QuantEcon/QuantEcon.jl/blob/master/REQUIRE#L1) to `julia 0.5`.
* [ ] `test.jl` should be removed.

src/sampler.jl
```diff
@@ -0,0 +1,64 @@
...
+        C[:L][i, i] >= -ATOL1 - RTOL1 * C[:L][1, 1] ||
+            throw(ArgumentError(non_PSD_msg))
+    end
+
+    tril!(view(A, :, 1:r))
+    A[:, r+1:end] = 0
+    Q = A[p,p]
+    for j in 1:n
+        for i in 1:n
+            isapprox(Q*Q', Sigma; rtol=RTOL2, atol=ATOL2) ||
```
oyamad: If you use `isapprox`, you don't have to do it n x n times, but just once.

Shunsuke-Hori:
Thank you @sglyon and @oyamad, 

It's done

oyamad:
@Shunsuke-Hori Many thanks, great work!
Merging.

PR #158, opened 2017-05-01
BUG: Fix MVNSampler

oyamad:
I found my code contained a bug...

```jl
n = 2
mu = zeros(n)
Sigma = eye(Int, n)
MVNSampler(mu, Sigma)
```

```
ERROR: ArgumentError: generic pivoted Cholesky factorization is not implemented yet
 in QuantEcon.MVNSampler{TM<:Real,TS<:Real,TQ<:Union{Float32,Float64}}(::Array{Float64,1}, ::Array{Int64,2}) at /Users/oyama/.julia/v0.5/QuantEcon/src/sampler.jl:26
```

### DIFF
--- a/src/sampler.jl
+++ b/src/sampler.jl
@@ -16,8 +16,10 @@ function MVNSampler{TM<:Real,TS<:Real}(mu::Vector{TM}, Sigma::Matrix{TS})
 
     n = length(mu)
 
-    if size(Sigma) != (n,n) # Check Sigma is n x n
-        throw(ArgumentError("Sigma must be 2 dimensional and square matrix of same length to mu"))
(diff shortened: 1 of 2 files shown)


sglyon:
As far as I can tell there are two changes here:

1. Use `cholfact` instead of `cholfact!`
2. Set `A = C.factors` instead of `A = cholfact!((copy(Sigma), :L), Val{true})`

Is that right?

oyamad:
Yes.
The eltype of the first argument of `cholfact!` must be `Float64` or `Float32`, but the previous code failed to convert the eltype. Now conversion is done inside `cholfact`.

PR #161, opened 2017-07-10
Discrete Markov chain estimation

cc7768:
Threw this together for kicks and giggles last night (it was a 15 minute exercise) in response to the [discourse thread](http://discourse.quantecon.org/t/estimate-markovchain-from-data-using-ml/239/2) by @jstac 

Uses maximum likelihood estimation to get transition matrix for a discrete number of states.

Attn: @jstac @sglyon

### DIFF
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -164,7 +164,7 @@ function estimate_MC_discrete{T}(X::Vector{T})
     state_i = findfirst(states, X[1])
     for t in 1:capT-1
         # Find next period's state
-        state_j = findfirst(states, X[t+1])
+        state_j = searchsortedfirst(states, X[t+1])
         cm[state_i, state_j] += 1.0
 
         # Tomorrow's state is j
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -180,7 +180,7 @@ function estimate_mc_discrete{T}(X::Vector{T}, states::Vector{T})
 
     # Make sure all of the passed in states appear in X... If not
     # throw an error
-    if any(map(x->in(x, X), states) .== false)
+    if any(!in(x, X) for x in states)
         error("One of the states does not appear in history X")
     end
 
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -155,6 +155,11 @@ Note: Because of the estimation procedure used, only states that are observed
 in the history appear in the estimated Markov chain... It can't divine whether
 there are unobserved states in the original Markov chain.
 
+For more info, refer to:
+
+- http://www.stat.cmu.edu/~cshalizi/462/lectures/06/markov-mle.pdf
+- https://stats.stackexchange.com/questions/47685/calculating-log-likelihood-for-given-mle-markov-chains
+
 ##### Arguments
 
 - `X::Vector{T}` : Simulated history of Markov states
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -134,10 +134,26 @@ end
 
 """
 Accepts the simulation of a discrete state Markov chain and estimates
-the transition probabilities via maximum likelihood.
+the transition probabilities
 
-Note: Only states that are observed in the history will be included
-in the estimation.
+Let S = {s₁, s₂, ..., sₙ} with s₁ < s₂ < ... < sₙ be the discrete
+states of a Markov chain. Furthermore, let P be the corresponding
+stochastic transition matrix.
+
+Given a history of observations, {X} where xₜ ∈ S ∀ t, we would like
+to estimate the transition probabilities in P, pᵢⱼ. For xₜ=sᵢ and xₜ₋₁=sⱼ,
+let P(xₜ | xₜ₋₁) be the pᵢⱼ element of the stochastic matrix. The likelihood
+function is then given by
+
+    L({X}ₜ; P) = P(x_1) ∏_{t=2}^{T} P(xₜ | xₜ₋₁)
+
+The maximum likelihood estimate is then just given by the number of times
+a transition from sᵢ to sⱼ is observed divided by the number of times
+sᵢ was observed.
+
+Note: Because of the estimation procedure used, only states that are observed
+in the history appear in the estimated Markov chain... It can't divine whether
+there are unobserved states in the original Markov chain.
 
 ##### Arguments
 
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -132,6 +132,10 @@ function _rouwenhorst(p::Real, q::Real, m::Real, Δ::Real, n::Integer)
 end
 
 
+# These are to help me order types other than vectors
+@inline _emcd_lt{T}(a::T, b::T) = isless(a, b)
+@inline _emcd_lt{T}(a::Vector{T}, b::Vector{T}) = Base.lt(Base.Order.Lexicographic, a, b)
+
 """
 Accepts the simulation of a discrete state Markov chain and estimates
 the transition probabilities
@@ -175,7 +179,7 @@ function estimate_MC_discrete{T}(X::Vector{T})
     capT = length(X)
 
     # Find all unique observations and sort them.
-    states = sort!(unique(X))
+    states = sort!(unique(X); lt=_emcd_lt)
     nstates = length(states)
 
     # Counter matrix
@@ -185,7 +189,7 @@ function estimate_MC_discrete{T}(X::Vector{T})
     state_i = findfirst(states, X[1])
     for t in 1:capT-1
         # Find next period's state
-        state_j = searchsortedfirst(states, X[t+1])
+        state_j = searchsortedfirst(states, X[t+1]; lt=_emcd_lt)
         cm[state_i, state_j] += 1.0
 
         # Tomorrow's state is j
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -174,12 +174,17 @@ For more info, refer to:
 transition matrix
 
 """
-function estimate_mc_discrete{T}(X::Vector{T})
+function estimate_mc_discrete{T}(X::Vector{T}, states::Vector{T})
     # Get length of simulation
     capT = length(X)
 
-    # Find all unique observations, sort them, and put into dictionary
-    states = sort!(unique(X); lt=_emcd_lt)
+    # Make sure all of the passed in states appear in X... If not
+    # throw an error
+    if any(map(x->in(x, X), states) .== false)
+        error("One of the states does not appear in history X")
+    end
+
+    # Count states and store in dictionary
     nstates = length(states)
     d = Dict{T, Int}(zip(states, 1:nstates))
 
@@ -203,3 +208,9 @@ function estimate_mc_discrete{T}(X::Vector{T})
     return MarkovChain(P, states)
 end
 
+function estimate_mc_discrete{T}(X::Vector{T})
+    # Get unique states and sort them
+    states = sort!(unique(X); lt=_emcd_lt)
+
+    return estimate_mc_discrete(X, states)
+end
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -130,3 +130,50 @@ function _rouwenhorst(p::Real, q::Real, m::Real, Δ::Real, n::Integer)
         return linspace(m-Δ, m+Δ, n), θN
     end
 end
+
+
+"""
+Accepts the simulation of a discrete state Markov chain and estimates
+the transition probabilities via maximum likelihood.
+
+Note: Only states that are observed in the history will be included
+in the estimation.
+
+##### Arguments
+
+- `X::Vector{T}` : Simulated history of Markov states
+
+##### Returns
+
+- `mc::MarkovChain{T}` : A Markov chain holding the state values and
+transition matrix
+
+"""
+function estimate_MC_discrete{T}(X::Vector{T})
+    # Get length of simulation
+    capT = length(X)
+
+    # Find all unique observations and sort them.
+    states = sort!(unique(X))
+    nstates = length(states)
+
+    # Counter matrix
+    cm = zeros(nstates, nstates)
+
+    # Compute conditional probabilities for each state
+    state_i = findfirst(states, X[1])
+    for t in 1:capT-1
+        # Find next period's state
+        state_j = findfirst(states, X[t+1])
+        cm[state_i, state_j] += 1.0
+
+        # Tomorrow's state is j
+        state_i = state_j
+    end
+
+    # Compute probabilities using counted elements
+    P = cm ./ sum(cm, 2)
+
+    return MarkovChain(P, states)
+end
+
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -174,22 +174,23 @@ For more info, refer to:
 transition matrix
 
 """
-function estimate_MC_discrete{T}(X::Vector{T})
+function estimate_mc_discrete{T}(X::Vector{T})
     # Get length of simulation
     capT = length(X)
 
-    # Find all unique observations and sort them.
+    # Find all unique observations, sort them, and put into dictionary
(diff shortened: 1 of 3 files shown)


src/markov/markov_approx.jl
```diff
@@ -130,3 +130,50 @@ function _rouwenhorst(p::Real, q::Real, m::Real, Δ::Real, n::Integer)
...
+    nstates = length(states)
+
+    # Counter matrix
+    cm = zeros(nstates, nstates)
+
+    # Compute conditional probabilities for each state
+    state_i = findfirst(states, X[1])
+    for t in 1:capT-1
+        # Find next period's state
+        state_j = findfirst(states, X[t+1])
```
sglyon: Is there a `findfirstsorted` or something like that. It seems kinda like a shame to have it sorted and then not leverage that?

test/test_markov_approx.jl
```diff
@@ -16,4 +16,27 @@
...
+    mc2 = estimate_MC_discrete(X)
+    @test isapprox(mc.state_values, mc2.state_values, atol=1e-10)
+    @test isapprox(mc.p, mc2.p, atol=1e-2)
+
+    # Test with other inputs
+    mc = MarkovChain(P, ["a", "b", "c"])
+    X = simulate(mc, 100_000)
+    mc2 = estimate_MC_discrete(X)
+    @test isapprox(mc.state_values, mc2.state_values, atol=1e-10)
+    @test isapprox(mc.p, mc2.p, atol=1e-2)
```
sglyon: A test where only a subset of states appears in the simulated path would be great.

Something like 


```
P = [1 0 0; 0.5 0.5 0; 1/3 1/3 1/3]
mc = MarkovChain(P, [10, 20, 30])
X = simulate(mc, 100_000, init=1)
```

Then the result should only be 2x2 and have values [10, 20]

sglyon:
Also see the travis error: https://travis-ci.org/QuantEcon/QuantEcon.jl/jobs/251842204#L334-L347

Looks like you can't set `atol` and not `rtol`?

test/test_markov_approx.jl
cc7768: Good thought. This should be tested.

I had originally planned on piggy backing on the other tests in `test_markov_approx.jl`, but when `n` got very large the simulations didn't cover all of the states in Rouwenhorst method so I decided just to write some smaller examples.

Done.

cc7768:
That's a weird error... I can run these on my computer :thinking:

src/markov/markov_approx.jl
cc7768: I could use `searchsortedfirst` which "Returns the index of the first value in `a` greater than or equal to `x`." I preferred `findfirst` because it is only checking for equality. This means that if it doesn't find the value then the index returned will be 0 and it will throw an error when it tries to use it... I'd prefer it to throw this error than silently just take the next largest value.

Unless the number of states you have is very big then I suspect this isn't a relevant performance hit -- And if you have that many states then you should be worried about having a long enough simulation to generate accurate transition matrices.

cc7768:
Fixed. It was trying to compute `rtol` for strings which it didn't know how to do... I've changed it to check for equality on state space since it is just copying those numbers anyways.

src/markov/markov_approx.jl
sglyon: I think you are ok with `searchsortedfirst` here because you _know_ any `X[t+1]` that can come up appears exactly once in `states` 

But, I take your point. With the small examples we have in the tests I would assume there is no performance difference.

cc7768: Ah. Good point. We know it will evaluate an equality on one of them.

Change made.

src/markov/markov_approx.jl
```diff
@@ -130,3 +130,50 @@ function _rouwenhorst(p::Real, q::Real, m::Real, Δ::Real, n::Integer)
         return linspace(m-Δ, m+Δ, n), θN
     end
 end
+
+
+"""
+Accepts the simulation of a discrete state Markov chain and estimates
+the transition probabilities via maximum likelihood.
```
sglyon: Could you add more description of the likelihood function you are using when you construct `p`? That would be great as a reference and docs.

jstac: There's a nice reference here:

http://www.stat.cmu.edu/~cshalizi/462/lectures/06/markov-mle.pdf

It's a nice and often cited result that max likelihood estimation of MC transition probabilities reduces to setting p_{ij} equal to the empirical frequency of transitions from i to j in the data.  This reference is also good:

https://stats.stackexchange.com/questions/47685/calculating-log-likelihood-for-given-mle-markov-chains

jstac:
@cc7768 Thanks!  It's really nice code too.  Very pretty.

@oyamad You might be interested.

jstac:
It might be worth mentioning in the docs that this is for scalar data.  If there's demand, we could probably make this work for more abstract state spaces later on.  (It would be a matter of replacing the observed states (which could, for example, be vector valued, or strings) with indices (label them `1,...,n`) and then proceeding as before.)

src/markov/markov_approx.jl
cc7768: I've added a description of the likelihood -- It was a bit of a pain to do with unicode, but I think I did about as good as unicode will let me. (Julia docs don't support mathjax/latex, right?)

Thanks for links John. I've added those to documentation to point people in right direction if they want more info.

cc7768:
This should be able to handle whatever kind of state you'd like to give it... It is parameterized by `T`. The requirement is that the simulation is a vector of "something" (which can include a vector of vectors).

Note: I thought I would get this for free, but since I now `sort` the states and use `searchsortedfirst` I have to take a stand on how vectors are ordered (I chose Lexicographically). Now you get it for free:

```
julia> P = random_stochastic_matrix(2)
2×2 Array{Float64,2}:
 0.13964   0.86036 
 0.537647  0.462353

julia> mc = MarkovChain(P, [[1.0, 2.0], [1.0, 3.0]])
Discrete Markov Chain
stochastic matrix of type Array{Float64,2}:
[0.13964 0.86036; 0.537647 0.462353]

julia> X = simulate(mc, 10_000);

julia> 

julia> mc2 = estimate_MC_discrete(X)
Discrete Markov Chain
stochastic matrix of type Array{Float64,2}:
[0.13609 0.86391; 0.534315 0.465685]

julia> mc2.state_values
2-element Array{Array{Float64,1},1}:
 [1.0, 2.0]
 [1.0, 3.0]
```

jstac:
@cc7768 Very nice ;-)

sglyon:
Great work!

Could you please add one test for the vector of vector case so that we make sure that works. After that I think this is ready to merge!

oyamad:
@cc7768 Good work!

I have some minor comments:

1. For lookup for state values I wonder which of `Dict` (or `OrderedDict`) and `searchsortedfirst` for a sorted array is faster. The latter should be O(log(n)), and the internet says dictionary is O(1) (I don't have much knowledge on data structures though).

2. There might be cases where the user has a state space with a preferred order. For such cases it might be useful to offer `estimate_MC_discrete(X, states)` where `states` is `Vector{T}` or `OrderedDict{T}`. Then it is likely that some states do not appear in `X`, for which one may assign arbitrary probabilities (say 1/n).

3. Is there any convention in using lower- or upper-case letters in the name of a function/method?

cc7768:
1. `Dict` is almost certainly a better way to do this. Great comment. I've made this change.
2. I'm not keen on arbitrarily assigning probabilities to the transition matrix, but if others think it is the right thing to do then I'm not strongly opposed to it. What do others think? @sglyon @jstac 
3. Yes, it looks like functions are typically all lower-case. I've renamed the function to reflect this.

jstac:
Regarding 2., I don't think we should allow the user to introduce states that are never visited in the data because, as @cc7768 says, there's no way to assign transition probabilities in a non-arbitrary manner.  

I guess though @oyamad wants to allow the user to introduce a particular ordering over the states, which sounds reasonable.

cc7768:
It now allows users to pass in states and throws an error if any of the states they specify are not observed in the history.

src/markov/markov_approx.jl
```diff
@@ -130,3 +130,87 @@ function _rouwenhorst(p::Real, q::Real, m::Real, Δ::Real, n::Integer)
...
+
+"""
+function estimate_mc_discrete{T}(X::Vector{T}, states::Vector{T})
+    # Get length of simulation
+    capT = length(X)
+
+    # Make sure all of the passed in states appear in X... If not
+    # throw an error
+    if any(map(x->in(x, X), states) .== false)
+        error("One of the states does not appear in history X")
```
sglyon: Sorry to be so picky, but I'm going to request one last change...

Will you please add a test that this error is thrown when we expect it to be thrown?

cc7768: Being picky now means that we spend less time maintaining the code later -- No apology is needed.

I've added a test for when it is given a state that doesn't show up in the history.

jstac:
@cc7768 Thanks!  Nice docs and nice code.

robertdkirkby:
Quick thoughts. Based on some experience doing this (approximating observations of a continuous variable with a discrete markov in a way that was free of distributional assumptions) as part of some value function iteration codes.

As I recall the MLE estimator is that once you have chosen your bins you then get the transition matrix by counting transitions and normalizing rows to (cumulative) sum to one. The trick is that to use this in practice you must first define the bins and I recall the MLE estimator saying nothing about how to do this (at least not in my short lit review). 

In my experience using this I found that a good approach was simply to borrow the automated bin choosing algorithms from histograms (I used whatever the algo that is the matlab default is). Pool all your observations, decide how many bins you want, and then apply the histogram algos to choose the bins. The intuition was that all the observations, when pooled, could be thought of as representing the stationary distribution. 

This worked well in practice and given historgram algos already exist is easy to implement. I tried a few alternatives to pick the bins (equally spaced points, points based on guassian quadrature that would be optimal if it were normal dist, etc.) but the historgram bins seemed to perform best. (Best here means when I graphed univariate time series those created with historgram bins looked more like original observed series to naked eye. One could obviously formalize a distance/loss criterion but I never bothered.) (The histogram algorithms can even suggest a number of bins as appropriate. Obviously based on different criteria to those you ideally want for markov chain. But can still be useful to give some idea that you might be using way too few to really capture the data.)

Note. This can also double as a method for discretizing any known process. Simulate your process and then apply this method. Likely you have already considered this as part of appeal but figured it might be worth a mention.

src/markov/markov_approx.jl
natashawatkins: > (Julia docs don't support mathjax/latex, right?)

It should: https://juliadocs.github.io/Documenter.jl/stable/man/latex/

@sglyon could we add support for this?

sglyon:
Hi @robertdkirkby thank you for the thoughtful message and the tips!

I believe (but please correct me if I am wrong @cc7768) that this  routine is meant to estimate the markov transition matrix for an observed _discrete_ process, rather than approximating a continuous variable as a discrete Markov process using observed data.

That being said, I think your suggestion is spot on and would be a useful addition to QuantEcon.jl. Would you be interested in contributing something like this here? If not, hopefully someone else on the team (or that sees this message -- consider this an open invitation to collaborate!) can find the time to do so.

src/markov/markov_approx.jl
sglyon: Hey @natashawatkins good catch.

This should already be ready to go as we are using Documenter to generate docs and math support seems to be enabled by default.

I think @Shunsuke-Hori  uses this in #163

cc7768:
Yes, this method is only for a discrete valued process, but the goal is to add some methods for continuous valued processes as well.

@robertdkirkby 's idea sounds really slick. One natural implementation of this (one of the reasons I think he mentioned it) is that the histogram method could build off of the discrete method I implemented -- First a "preparatory" set would discretize data using histograms and then pass it off to this method to estimate transitions.

Do you have some code (in any language) that does this? If so, (and as long as it is licensed under something that permits us to build on it) then I would love to see an example. It also sounds like what you're describing would allow bins of different sizes etc... I hadn't heard of this method but it seems like a good one -- I am keen for QE to add this functionality :)

robertdkirkby:
I did it as part of a replication of Hansen (1985). 
His shock process is yt=rho*yt-1+et, where et is log-normal.
[note difference from common logyt=rho*logyt-1+et, where et is normal]
He used linear-quadratic methods, so only conditional mean of shock matters and the 'log-normal' is essentially ignored. I found that difference was important for certain key moments of model
[Replication results form part of my paper: [https://link.springer.com/article/10.1007/s10614-015-9544-1](https://link.springer.com/article/10.1007/s10614-015-9544-1) ]

In my case goal was to get a discrete MC approximation of an unusual (known) shock process, but conceptually there is no reason you can't just do exactly the same from any observed time series data.

The matlab code I wrote is lines 51 to 83 of
[https://github.com/vfitoolkit/vfitoolkit-matlab-replication/blob/master/Hansen1985/Hansen1985.m](https://github.com/vfitoolkit/vfitoolkit-matlab-replication/blob/master/Hansen1985/Hansen1985.m)
But the relevant part is really just a single line to call the histogram command on line 74. And the lines after that which are my version of the command @cc7768 has just coded up here.
My code is unlikely to be useful to build on in sense that it is just a call to matlab histogram fn, together with the part you already coded, but does give a demo.

I took a further look in my archives and I did actually test the performance based on following: simulate standard AR(1) using a random number generator and monte-carlo, use historgram-plus-MLE-for-discrete-transition-matrix method to approximate it as a discrete MC, simulate discrete MC, estimate AR(1) from simulation. I found that as I increased number of grid pts you got back same AR(1) that you put in.

As @cc7768 observes, it allows for bins of different sizes. One can loosely justify the process by appealing to decomposition of a markov process into pairing of a copula (transition function on rank) and stationary distribution. The histograms are well understood non-parametric estimators for the stationary distribution based on bandwidth estimators with data-selected-bandwidth. The transition matrix is then also a well understood estimator: MLE for transition of discrete markov. The informal part is their joint use, theoretically you could obviously somehow do better by doing both jointly. I did not find anyone who had sat down and done this theory though. Advantage of this approach is the ease of implementation since histogram commands are standard in statistics packages. For your purposes I expect there are some standard historgram commands in python/julia stats packages that could be repurposed.

robertdkirkby:
Took a look at code:  src/markov/markov_approx.jl
In line 195 you start from d[X[1]]. Your first step of for loop (lines 196 to 203) also starts from t=1 and d[X[t]]. Unless I misinterpret you are counting the nontransition from t=1 to t=1 and then missing the final transition from t=capT-1 to t=capT.

Solution would just be to change line 196 to: for t in 2:capT

I don't have copy of Julia and QuantEcon on this computer, otherwise would set this up as pull request myself.

cc7768:
@robertdkirkby Good catch -- Thanks for noticing that!

I've changed it to `state_j = d[X[t+1]]`

@sglyon -- Just a heads up that since this was so minor I pushed directly to master.

jstac:
@robertdkirkby Regarding your comment

> In my experience using this I found that a good approach was simply to borrow the automated bin choosing algorithms from histograms (I used whatever the algo that is the matlab default is). Pool all your observations, decide how many bins you want, and then apply the histogram algos to choose the bins. The intuition was that all the observations, when pooled, could be thought of as representing the stationary distribution.

I think this is a great idea.  My loose idea was to resample from the ergodic distribution to obtain points and then use some kind of nearest neighbors algorithm to turn these points into regions, which correspond to your bins.  But your idea makes just as much sense, if not more, and, as you say, the code for generating bins is already there.

There's a paper waiting to be written here, showing that the method can be used to approximate any time homogenous Markov process in R^n.  Moreover, it's very simple and flexible.

One might argue that a simulation based method is inefficient, but this is hardly an issue, because the discretization is done once off.   Moreover, if the n in R^n is relatively large (even 2, 3 and up), Monte Carlo might in fact be an efficient option.

I don't have time to write such a paper but I'd love to see this code in QuantEcon either way.  As @cc7768 says,

> the histogram method could build off of the discrete method I implemented -- First a "preparatory" set would discretize data using histograms and then pass it off to this method to estimate transitions.

@robertdkirkby Your comments are much appreciated.  The hive mind turns out to be efficient ;-)

CC @cc7768 @sglyon @oyamad @vgregory757

PR #163, opened 2017-07-11
Add discreteVAR of Farmer and Toda

Shunsuke-Hori:
This PR adds `discreteVAR` by Farmer and Toda, which was proposed by [the discourse's post](http://discourse.quantecon.org/t/alternative-method-for-discretizing-general-state-markov-processes/233/6)

The codes, documentations, and tests are borrowed from their [matlab files](https://sites.google.com/site/discretevar/). 

This PR only focuses on evenly spaced grid method.


jstac:
@Shunsuke-Hori Thanks!!  I guess this was a lot of work.  How did output compare to the Matlab code?

Shunsuke-Hori:
Measured by sup norm, the difference of outputs for each test is
- test 1: AR(1)
    - p: 8.20586931737921e-9
    - state_values: 4.618527782440651e-14
- test 2: VAR(1)
    - p: 4.0857455499421214e-7
    - state_values: 1.807449843960196e-7

cc7768:
This looks very cool. Thanks for doing this!

I'm hoping to read this a bit more carefully in next few days, but I'm wondering why you need both `NLopt` and `Optim`? Could we make do with one or the other to minimize dependencies?

Shunsuke-Hori:
@cc7768 There are two optimization problems: [one](https://github.com/QuantEcon/QuantEcon.jl/pull/163/files#diff-5f1f9f99a9b527e5dd7b3464fb689d22R404) using hessian and the [other one](https://github.com/QuantEcon/QuantEcon.jl/pull/163/files#diff-5f1f9f99a9b527e5dd7b3464fb689d22R534) of constrained maximization

I agree with you, but `Optim` does not support constrained optimization and `NLopt` does not support optimization with hessian

Shunsuke-Hori:
Reminder to me
- [ ] remove `RowVector`
- [ ] `state_values` should be `Vector` for AR(1) (instead of `Vector{Vector}` aimed for VAR)

sglyon:
Very cool @Shunsuke-Hori! Thank you for writing this up.

I see that you've given yourself a todo list here, which is great. Before I do a more thorough review (perhaps with the help of other QE members), I will let you work through your own list.

I do have one quick comment/suggestion. Could you use `lowercase_with_underscores` instead of `camelCase` for function names? That would be consistent with the rest QuantEcon.jl and the majority of Julia code. Thanks!

jstac:
@Shunsuke-Hori I recommend `discrete_var`, since it's more descriptive.

sglyon:
Thanks @jstac -- I recommend the same name.

src/markov/markov_approx.jl
```diff
@@ -214,3 +218,402 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+          1 5 7
+           ...
+          3 6 8
+          3 6 9]  # 27 x 3 array
+##### Arguments
+- `A::Matrix` : (N x M) Matrix
+##### Returns
+- `A_comb` : (N^M x M) Matrix, combination of each row of A
+"""
+function allcomb3(A::Matrix)
```
cc7768: Is there a reason you need the first column to be the slow moving column? If not, you might consider defining an even simpler function (thanks @sglyon for idea)
```
allcomb3(A::Matrix) = gridmake([view(A, :, i) for i in 1:size(A, 2)]...)
```
Or we could add a new method for `gridmake` -- I think @sglyon is a little hesitant on this because it is inconsistent with how we handle `gridmake(Vector, Matrix)`. See example below:

```
julia> gridmake([1, 2], [10 20; 30 40])
4×3 Array{Int64,2}:
 1  10  20
 2  10  20
 1  30  40
 2  30  40
```
whereas new method would do
```
julia> gridmake([10 20; 30 40])
4×2 Array{Int64,2}:
 10  20
 30  20
 10  40
 30  40
```

src/markov/markov_approx.jl
```diff
@@ -214,3 +218,402 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    return obj
+end
+
+"""
+Compute gradient of objective function
+##### Returns
+- `grad` : (L x 1) gradient vector of the objective function evaluated
+           at lambda
+"""
+function entropy_grad!(lambda::Vector, grad::Vector, Tx::Matrix, TBar::Vector, q::Matrix)
```
cc7768: Julia convention is typically to put the mutated arguments first -- Could we make `grad` the first argument?

src/markov/markov_approx.jl
```diff
@@ -214,3 +218,402 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    temp2 = temp.*Tdiff
+    grad .= vec(sum(temp2, 2))
+end
+
+"""
+Compute hessian of objective function
+##### Returns
+- `hess` : (L x L) hessian matrix of the objective function evaluated at `lambda`
+"""
+function entropy_hess!(lambda::Vector, hess::Matrix, Tx::Matrix, TBar::Vector, q::Matrix)
```
cc7768: See comment about mutated arguments -- Could we put `hess` first?

src/markov/markov_approx.jl
```diff
@@ -214,3 +218,402 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+
+    K = size(A, 1) # size of A
+    d = trace(A)/K # diagonal of U'*A*U should be closest to d
+    function obj(X, grad)
+        X = reshape(X, K, K)
+        return (norm(diag(X'*A*X)-d))
+    end
+    function unitaryConstraint(res, X, grad)
+        X = reshape(X, K, K)
+        res .= vec(X'*X - eye(Int64(sqrt(length(X)))))
```
cc7768: Does this work on 0.5? It doesn't work on 0.6.

`Int64(sqrt(length(X)))` should either be `round(Int, sqrt(length(X)))`, `floor(Int, sqrt(length(X)))`, or `ceil(Int, sqrt(length(X)))` depending on the behavior you want.

Shunsuke-Hori:
@jstac @cc7768 @sglyon Thank you for the review and advices, they are really helpful!!

I'll reply to some comments.

src/markov/markov_approx.jl
```diff
@@ -214,3 +218,402 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+  dimensions due to curse of dimensionality issues.
+- Future updates will allow for singular variance-covariance matrices and
+  sparse grid specifications.
+"""
+function discrete_var{TI<:Integer}(b::ScalarOrArray, B::ScalarOrArray,
+                     Psi::ScalarOrArray, Nm::TI,
+                     nMoments::TI=2, method::Symbol=:even,
+                     nSigmas::Union{Void, TI}=nothing)
+
+    if typeof(B) <: Real
```
Shunsuke-Hori: @cc7768 I tried to do that before but it was not straightforward for some reason (which I forgot). Let me carefully work on the stuff regarding dispatching next week.

Another question I have is efficiency. The benefit of following your strategy that scalar `b` is passed to a general function as `b=ones(M)*b` is we can avoid `if...else..` style branching. However, this makes all calculation `1x1` matrix operations instead of scalar ones. I am not sure but it might be faster to deal with scalar as scalar, because this function requires many matrix operations for VAR and `1x1` matrix operations are more time-consuming than scalar operations. Do you have any thoughts?

src/markov/markov_approx.jl
Shunsuke-Hori: > Is there a reason you need the first column to be the slow moving column?

Yes. Otherwise, rows and columns of a lot of matrices must be reordered since the output of `allcomb3` is used for matrix operations. Instead of `allcomb3(A::Matrix) = gridmake([view(A, :, i) for i in 1:size(A, 2)]...)`, I edited the code and wrote a shorter function.

src/markov/markov_approx.jl
Shunsuke-Hori: @cc7768 Instead of passing scalar as `ones(1,1)*b`, I dispatched scalar (AR(1)) case and vector (VAR(1)) case. 

I think it is faster to treat scalar as scalar than treat it as `1x1` matrix.

CC @sglyon , @jstac

Shunsuke-Hori:
@cc7768 @jstac @sglyon Following the review comments, I updated the code. Docs are also updated following #168 

By the way, the test fails due to the build error of `NLopt`. I have no idea why it fails. Could anyone help me to fix this?

Shunsuke-Hori:
I don't know why but the previous commit makes the test pass.

jstac:
@Shunsuke-Hori Thanks!  Nice work with the docs.

sglyon:
Thanks for the work @Shunsuke-Hori. 

Is this ready for another round of review? @cc7768  did a great job last time -- I'll take a look this time around whenever you are ready for me to go through it.

Shunsuke-Hori:
@sglyon Yes, I appreciate the second review.

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
 
     return estimate_mc_discrete(X, states)
 end
+
+"""
+
+types specifying the method for `discrete_var`
+
+"""
+@compat abstract type AbstractMethod end
```
sglyon: I think this should be more descriptive than just `AbstractMethod`. Maybe `VAREstimationMethod` (I think if we have something descriptive then we don't need the `Abstract`)

Open to other thoughts here.

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+- `Nm::Integer > 3` : Desired number of discrete points in each dimension
+
+##### Optional
+
+- `nMoments::Integer` : Desired number of moments to match. The default is 2.
+- `method::AbstractMethod` : AbstractMethod specifying the method used to determine
+              the grid points. Accepted inputs are `Even()`. Please see the
+             paper for more details.
+             NOTE: `Quantile()` and `Quadrature()` are not supported now.
+- `nSigmas::Real` : If the `Even()` option is specified, nSigmas is used to
```
sglyon: Do these docs render properly with the spacing of lines 2+ in bullet items?

If so, great. I think what you've written is readable in plain text. As long as it renders that'd be great!

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+              determine the number of unconditional standard deviations
+              used to set the endpoints of the grid. The default is
+              sqrt(Nm-1).
+
+##### Returns
+
+- `P` : (Nm^M x Nm^M) probability transition matrix. Each row
```
sglyon: Do these need math backticks?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+        corresponds to a discrete conditional probability
+        distribution over the state M-tuples in X
+- `X` : (M x Nm^M) matrix of states. Each column corresponds to an
+        M-tuple of values which correspond to the state associated
+        with each row of P
+
+##### NOTES
+
+- discrete_var only accepts non-singular variance-covariance matrices.
```
sglyon: Let's put this comment up where Psi is documented.

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+##### Arguments
+
+- `x::Real` : Scalar value
+
+##### Return
+
+- `(1, 1)`
+
+"""
+Base.size(x::Real) = (1, 1)
```
sglyon: We should **not** do this.

A general rule of thumb in Julia is to not add methods to a function you didn't create (e.g. something from a different module) on types you didn't define. Here we (QuantEcon.jl) doesn't own `Base.size` or `Real`, so we shouldn't define this method.

you probably did this to handle the scalar and matrix cases more seamlessly -- if so, let's see if we can find another way To do this.

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    # Check size restrictions on matrices
+    M == M_ || throw(ArgumentError("B must be a scalar or square matrix"))
+    size(b, 2) == 1 || throw(ArgumentError("b must be a column vector"))
+    M == length(b) || throw(ArgumentError("b must have the same number of rows as B"))
+
+    #% Check that Psi is a valid covariance matrix
+    isposdef(Psi) || throw(ArgumentError("Psi must be a positive definite matrix"))
+
+    # Check that Nm is a valid number of grid points
+    Nm>=3 || throw(ArgumentError("Nm must be a positive interger greater than 3"))
```
sglyon: Spacing around `>=`

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+  contains the same number of points. For this reason it is recommended
+  that this code not be used for problems of more than about 4 or 5
+  dimensions due to curse of dimensionality issues.
+- Future updates will allow for singular variance-covariance matrices and
+  sparse grid specifications.
+
+"""
+function discrete_var{TI<:Integer}(b::ScalarOrArray, B::ScalarOrArray,
+                     Psi::ScalarOrArray, Nm::TI,
+                     nMoments::TI=2, method::AbstractMethod=Even(),
```
sglyon: My preference would be to use `nmoments` or `n_moments` instead of `nMoments`.

Same applies to `nSigmas`.

Thoughts from the team?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+- discrete_var only accepts non-singular variance-covariance matrices.
+- discrete_var only constructs tensor product grids where each dimension
+  contains the same number of points. For this reason it is recommended
+  that this code not be used for problems of more than about 4 or 5
+  dimensions due to curse of dimensionality issues.
+- Future updates will allow for singular variance-covariance matrices and
+  sparse grid specifications.
+
+"""
+function discrete_var{TI<:Integer}(b::ScalarOrArray, B::ScalarOrArray,
```
sglyon: The `TI` parameter doesn't seem necessary to me. The role it plays here is to restrict `Nm` and `nMoments` to have the same type. I don't think we need this.

What about writing it as `Nm::Integer, nMoments::Integer`.

I believe the only place `TI` is used in the body of the function is when defining the upper bound on the loop over `k`. I think we can probably just replace that with `Int` as it doesn't seem critical for that to match the type of `Nm` or `nMoments`

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    M == length(b) || throw(ArgumentError("b must have the same number of rows as B"))
+
+    #% Check that Psi is a valid covariance matrix
+    isposdef(Psi) || throw(ArgumentError("Psi must be a positive definite matrix"))
+
+    # Check that Nm is a valid number of grid points
+    Nm>=3 || throw(ArgumentError("Nm must be a positive interger greater than 3"))
+
+    # Check that nMoments is a valid number
+    if nMoments < 1 || !(nMoments%2 == 0 || nMoments == 1)
```
sglyon: Space around `%`

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+                    temp[jj, :], _, _ = discrete_approximation(y1D[jj, :],
+                        X -> (X-condMean[jj,ii])/scalingFactor[jj],
+                        0, reshape(q[jj, :], 1, Nm), 0)
+                    lambdaBar[(jj-1)*2+1:jj*2, ii] = zeros(2,1)
+                elseif nMoments == 2
+                    lambdaBar[(jj-1)*2+1:jj*2, ii] = lambda
+                    temp[jj, :] = p
+                else # solve maximum entropy problem sequentially from low order moments
+                    lambdaBar[(jj-1)*2+1:jj*2, ii] = lambda
+                    for mm = 4:2:nMoments
```
sglyon: What happens in the  `nMoments == 3` case?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+                        end
+                    end
+                    temp[jj, :] = p
+                end
+            end
+        end
+        P[ii, :] .= vec(prod(allcomb3(temp'), 2))
+    end
+
+    X = C*D + repmat(mu, 1, Nm^M) # map grids back to original space
```
sglyon: I think you can use broadcasting instead of `repmat` here

Does 

```julia
X = C*D .+ mu'
```

give you the right answer?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+
+    M != 1 || (return MarkovChain(P, vec(X)))
+    return MarkovChain(P, [X[:, i] for i in 1:Nm^M])
```
sglyon: This is more of a design question than anything else. Should we have state values be `Vector{Vector{T}}` or should it be `Vector{NTuple{M,T}}`?

I'd love to hear input from others on this one

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+- `method::Even` : method for grid making
+
+##### Return
+
+- `y1D` : (M x Nm) matrix of variable grid
+
+"""
+function construct_1D_grid(Sigma::ScalarOrArray, Nm::Integer,
+                           M::Integer, nSigmas::Real, method::Even)
+    if M == 1
```
sglyon: I _believe_ we can drop the conditional on `M` here by doing 

```julia
minSigmas = sqrt(minimum(eigfact(Sigma).values))
```

`eigfact` will work for both scalars and matrices

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+##### Returns
+
+- `p` : (1 x N) vector of probabilties assigned to each grid point in `D`.
+- `lambdaBar` : (L x 1) vector of dual problem variables which solve the
+                maximum entropy problem
+- `momentError` : (L x 1) vector of errors in moments (defined by moments of
+                  discretization minus actual moments)
+
+"""
+function discrete_approximation(D::Vector, T::Function, TBar::Vector,
```
sglyon: Can we loosen `Vector` and `Matrix` to be their `AbstractXXX` counterparts, or do rely on an actual array?

If so, let's do this for all the functions below also

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+                                  q::Matrix=ones(N)'/N, # Default prior weights
```
sglyon: My vote would be to have `q` input as an `AbstractVector` of length `N` instead of a `1 x N` matrix. 

I think it is easier (for me at least) to just pass a vector and then let you reshape it inside the function. 

Same holds for `lambda0`

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+Compute a discrete state approximation to a distribution with known moments,
+using the maximum entropy procedure proposed in Tanaka and Toda (2013)
+
+```julia
+p, lambdaBar, momentError = discrete_approximation(D, T, TBar, q, lambda0)
+```
+
+##### Arguments
+
+- `D::Vector` : (N x 1) vector of grid points. K is the dimension of the
```
sglyon: Let's drop the ` x 1` here. I think "length `N`" fits better with how Julia arrays are represented and talked about.

Also, where does the `K` come from that you refer to on this line?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+        domain. N is the number of points at which an approximation
+        is to be constructed.
+- `T::Function` : A function handle which should accept arguments of dimension
```
sglyon: If the function `T` needs to accept a 1d array, let's change this language to

```
- `T::Function` : A function that accepts a single `AbstractVector` of length N and returns an (L x N) ...
```

(the `...` means to keep what you had after that)

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+function entropy_obj(lambda::Vector, Tx::Matrix, TBar::Vector, q::Matrix)
+
+    # Some error checking
+    L, N = size(Tx)
+
+    !(length(lambda) != L || length(TBar) != L || length(q) != N) ||
+        error("Dimensions of inputs are not compatible.")
+
+    # Compute objective function
+    Tdiff = Tx-repmat(TBar, 1, N)
```
sglyon: I think we should be able to do `Tdiff = Tx .- Tbar` here

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+- `q::Matrix` : (1 X N) vector of prior weights for each point in the grid.
+
+##### Returns
+
+- `obj` : scalar value of objective function evaluated at lambda
+
+"""
+function entropy_obj(lambda::Vector, Tx::Matrix, TBar::Vector, q::Matrix)
+
+    # Some error checking
```
sglyon: Is `entropy_obj` something users will call by themselves, or just a internal routine we use in `discrete_approximation`?

If it is meant to be internal, could we do the error checking just once from within `discrete_approximation` so ti doesn't happen every iteration of the solver?

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,567 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    obj = sum(temp)
+    return obj
+end
+
+"""
+Compute gradient of objective function
+
+##### Returns
+
+- `grad` : (L x 1) gradient vector of the objective function evaluated
```
sglyon: Let's drop the `x 1`

src/markov/markov_approx.jl
cc7768: I agree that more descriptive is always better.

src/markov/markov_approx.jl
Shunsuke-Hori: It's invalid number: see [this test](https://github.com/QuantEcon/QuantEcon.jl/pull/163/files#diff-b49beb3db45370dfe66a3118785f6919R105)

src/markov/markov_approx.jl
Shunsuke-Hori: Yes
docs

src/markov/markov_approx.jl
jstac: agree

src/markov/markov_approx.jl
Shunsuke-Hori: @sglyon How about defining another function like this? Is this also what we shoud avoid?

~~~
size_scalar_as_matrix(x::Real) = (1, 1)
size_scalar_as_matrix(x::Matrix) = size(x)
~~~

src/markov/markov_approx.jl
Shunsuke-Hori: @sglyon Since math-rendered x looks differently, I avoid math backticks.

I added code-backticks to make it consistent with other existing docs.

src/markov/markov_approx.jl
sglyon: That would be ok as we would be the ones to define or own the function `size_scalar_as_matrix`

I wonder if there is a cleaner way to do this leveraging multiple dispatch -- meaning to dispatch on the types at an earlier stage than when you need the size.

Another alternative would be to use

```
size(x, 1)
```

and

```
size(x, 2)
```

as that will work for scalar and array types in Julia (for numbers `size(x, N::Int)` always returns 1)

Shunsuke-Hori: @sglyon 
> I wonder if there is a cleaner way to do this leveraging multiple dispatch 

I could not come up with that. `size(x, 1)` and `size(x, 2)` seem nice. Thank you!!

src/markov/markov_approx.jl
Shunsuke-Hori: This design issue remains untouched. 

Are the team members fine with `Vector{Vector{T}}`?

jstac: `Vector{NTuple{M,T}}` seems nicer to me, since it's more explicit, but I don't have a strong opinion.

sglyon: I also like the idea of that, but the downside is that `Vector{Tuple{M,T}}` is not type stable (`M` is not a compile time value).

There are some tricks we can do to make that happen, but it might not be worth it...

cc7768: I don't mind the `Vector{Vector{T}}` too much. If the relevant output is something like `dot(e, state)` then `Vector{Vector{T}}` might even be better (think Bansal Yaron type models where consumption growth is a sum of some of the state variables). I guess the downside is that there is nothing forcing the vectors to be the same length.

Shunsuke-Hori: I don't think we can directly apply `dot` because the inner `Vector` collects one of the states of each variable. For example, if the model is a VAR of two variables, ``x`` and ``y``, the inner vector is `[x_1, y_1]` but not `[x_1, x_2, ... ]`. If you really want to apply `dot` directly, `Matrix` would work but I know it is inconsistent and `MarkovChain` does not accept that.

cc7768: What I have in mind is there are N m-dimensional states, `X = {x_i \in R^m}_{i=1}^{N}`. For a given vector `e` and current `x_i \in X`, a relevant object in the model might be `y = dot(e, x_i)`. If `x_i` is stored as a tuple then we can't compute `y` using `dot(e, x_i)` but if it is a vector then we can -- In your notation, what I need is the vector `[x_1, y_1]`.

sglyon: Let's keep the vector of vector

Shunsuke-Hori: Ah, I see!! that's surely a possible usage. Thank you for the clarification. I agree with @cc7768 and vote to `Vector{Vector{T}}`.

sglyon:
Thank you for responding to my requests.

Do you feel this is ready for a (hopefully) final review?

Shunsuke-Hori:
Yes, please review

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,543 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+                      Nm::Integer,
+                      n_moments::Integer=2,
+                      method::VAREstimationMethod=Even(),
+                      n_sigmas::Real=sqrt(Nm-1))
+
+    M, M_ = size(B, 1), size(B, 2)
+
+    # Check size restrictions on matrices
+    M == M_ || throw(ArgumentError("B must be a scalar or square matrix"))
+    size(b, 2) == 1 || throw(ArgumentError("b must be a column vector"))
```
sglyon: This check is not necessary because we only allow scalar or AbstractVector in the type constraint on the function argument

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,543 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+    # Check that Nm is a valid number of grid points
+    Nm >= 3 || throw(ArgumentError("Nm must be a positive interger greater than 3"))
+
+    # Check that n_moments is a valid number
+    if n_moments < 1 || !(n_moments % 2 == 0 || n_moments == 1)
+        error("n_moments must be either 1 or a positive even integer")
+    end
+
+    # Compute polynomial moments of standard normal distribution
+    gaussianMoment = zeros(n_moments)
```
sglyon: I know these are internal variable names, but would you please change them to have underscores instead of camel case?

In this case I'm asking for the variable to be `gaussian_moment`.

There are a few others throughout the code in this PR, but I'll only comment once to avoid excessive emails/comments.

In general the rule of thumb is to use `UpperCamelCase` for type names (including abstract, struct, and mutable struct) and use` lowercase_with_underscores` for variable and function names. I realize I wasn't ever clear about that before -- sorry.

Thanks!

Shunsuke-Hori:
Thank you for the review. I updated the code.

One question: I realized the code does not include the reference to the original paper. What is the convention of julia package or rule of QE? Is it enough to add one comments like 

`This is developed by Farmer and Toda (Quantitative Economics, 2017)`

into the doc of `discrete_var`?

sglyon:
I think putting a reference in the docstring for the method would be good for now.

Shunsuke-Hori:
Thank you @sglyon. I added a reference to the original paper.

sglyon:
Hey @cc7768 if you have a chance to make sure this is ready for merge and then press the green button (squash and merge) in the next couple eof days that would be great.

Otherwise I'll get back to it probably on Saturday.

Thanks!

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,548 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+            # Try to use intelligent initial guesses
+            if ii == 1
+                lambda_guess = zeros(2)
+            else
+                lambda_guess = lambda_bar[(jj-1)*2+1:jj*2, ii-1]
+            end
+
+            # Maximum entropy optimization
+            if n_moments == 1 # match only 1 moment
+                temp[jj, :], _, _ = discrete_approximation(y1D[jj, :],
```
sglyon: Sorry to bring this up now, but is there a reason we _need_ temp to be ordered the way it is? The code might be more efficient if it was laid out in column major order (e.g. using `temp[:, jj]` instead of `temp[jj, :]`). 

I don't think this will make a big difference, especially relative to the cost of calling solvers, so if you have a reason for the current ordering I won't push too hard on this.

src/markov/markov_approx.jl
```diff
@@ -219,3 +223,548 @@ function estimate_mc_discrete{T}(X::Vector{T})
...
+
+"""
+
+Compute the moment defining function used in discrete_approximation
+
+```julia
+T = polynomial_moment(X, mu, scaling_factor, mMoments)
+```
+
+##### Argumentss:
```
sglyon: Spelling error

src/markov/markov_approx.jl
Shunsuke-Hori: Thank you for the repeated reviews. I was simply using the original matlab code without paying attention to that point. (matlab is also column major, though...) Fixed.

src/markov/markov_approx.jl
Shunsuke-Hori: Fixed.

sglyon:
Thank you!

I know this review process was quite long (much longer than average), but you were very effective at responding to reviews

PR #164, opened 2017-07-13
ENH: use rand instead of draw for drv sampling 

### LINKED ISSUE #162: Question: change `draw(::DiscreteRV)` to `rand(::DiscreteRV)`?
What do people think about changing the function name for getting a random sample from a DiscreteRV from `draw` to `rand`? The `rand` function seems to be the right name in the Julia ecosystem.

I'd also be open to having these two names be synonyms for one another.

cc @jstac

sglyon:
closes #162


jstac:
Looks good.

PR #165, opened 2017-07-14
add latex support to documenter

natashawatkins:
Hi @sglyon, this should add support for latex in Documenter.jl

Would you and @mmcky be able to test this?
@mmcky can you see if there is an equivalent of readthedocs to trigger a build of this branch of the documentation

Info is here: https://juliadocs.github.io/Documenter.jl/v0.1.3/

arma.jl docstring has been edited for testing

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -19,6 +19,6 @@ deploydocs(
     branch = "gh-pages",
     target = "build",
     julia  = "0.5",
-    deps = nothing,
+    deps = Deps.pip("pygments", "mkdocs", "python-markdown-math"),
     make = nothing,
 )
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -15,3 +15,10 @@ pages:
 - Overview: 'api/index.md'
 - API Docs:
   - QuantEcon: 'api/QuantEcon.md'
+
+markdown_extensions:
+  - mdx_math
+  
+extra_javascript:
+- https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML
+- assets/mathjaxhelper.js
--- a/src/arma.jl
+++ b/src/arma.jl
@@ -21,7 +21,9 @@ Represents a scalar ARMA(p, q) process
 If phi and theta are scalars, then the model is
 understood to be
 
-    X_t = phi X_{t-1} + epsilon_t + theta epsilon_{t-1}
+```math
+X_t = \phi X_{t-1} + \epsilon_t + \theta \epsilon_{t-1}
+```
 
 where epsilon_t is a white noise process with standard
 deviation sigma.


sglyon:
Hey @natashawatkins thanks for doing this!

A few things to note.

We use the native html output (see [here](https://juliadocs.github.io/Documenter.jl/stable/man/guide/#Output-formats-1)) so the mkdocs specific changes in this PR won't make a difference.

I think all that should be required is that we make the proper annotations to denote math blocks.

Also, documeter.jl will automatically push docs to http://quantecon.github.io/QuantEcon.jl/latest/ whenever we commit to master and travis-ci passes. You can verify this by seeing [docs for @cc7768  new routine for discrete mc estimation](http://quantecon.github.io/QuantEcon.jl/latest/api/QuantEcon.html#QuantEcon.estimate_mc_discrete-Tuple{Array{T,1},Array{T,1}})

sglyon:
Once travis passes on the push to master: https://travis-ci.org/QuantEcon/QuantEcon.jl/builds/253584266

let's check http://quantecon.github.io/QuantEcon.jl/latest/api/QuantEcon.html#QuantEcon.ARMA to see if the math is rendered properly

natashawatkins:
Thanks @sglyon  :) I'm happy to do the math block stuff once we confirm it works

Btw, is this because my computer is missing something related to unicode...
screen shot 2017-07-14 at 10 19 20 pm

natashawatkins:
It almost worked! Although looks like I didn't escape correctly..
screen shot 2017-07-14 at 11 00 32 pm

natashawatkins:
@sglyon Sorry, is there any way to build the docs locally? When i run `julia docs/make.jl` my changes aren't reflected in the `build` directory...

sglyon:
Hey @natashawatkins, running `julia docs/make.jl` should put the output in `docs/build` -- it works on my machine...

sglyon:
And I see the same thing in my browser with the unicode in the `estimate_mc_discrete` docstring. Now that we know we have math support we should switch to latex instead of unicode

natashawatkins:
@sglyon Yeah I'm not sure, it just seems to create an old version of the docs - Matt mentioned it could be because it's using the version of Julia installed on my computer?

sglyon:
Hmm, that doesn't sound right to me. I just tried with both Julia 0.5 and 0.6 and it worked without a problem.

Maybe try to fully delete the `docs/build` directory and try again?

Note that you need to start a simple web server to view the docs

natashawatkins:
I definitely removed the build folder, ran `julia docs/make.jl` and then started a served in the `build` directory with `python -m http.server`

And it still shows the old version (without even the incorrectly escaped formatting pushed before)

sglyon:
Hmm. that is very strange. Is it possible that you are checked out to a different branch that doesn't include the latest docstring updates?

Also, you are looking at the `docs/build/` directory, right? 

(Sorry -- I realize you've probably done all this correctly, I'm just trying to identify the issue)

natashawatkins:
That's ok - I'd like to know as well.

I'm in the branch in which I edited the docstrings.

And yes i'm opening `/docs/build/index.html`

sglyon:
Issue solved. The problem was working with this repo in a directory other than `~/.julia/v0.5/QuantEcon` (@natashawatkins is on Julia v0.5).

The reason that is a problem is that Documenter loads the QuantEcon module from the `LOAD_PATH` (e.g. folders in `~/.julia/v0.5/`) and extracts docstrings from the loaded module. Only changes to docstrings in that location are visible to documenter and put in the resulting html

PR #166, opened 2017-07-14
add latex rendering in arma.jl

natashawatkins:
* This PR adds latex rendering to `arma.jl`


sglyon:
Looks great! 👍 

Thanks

PR #167, opened 2017-07-14
Allow for noisy observations in LSS class

vgregory757:
Added the H matrix like [here](https://lectures.quantecon.org/jl/linear_models.html#noisy-observations). 

It's optional with default argument 0, so it shouldn't change any existing code.


jstac:
@vgregory757 Thanks! It's great to have this functionality and the code looks good.

PR #168, opened 2017-07-15
Latex rendering in docs + other improvements

natashawatkins:
Yay it's pretty now

* added latex rendering to all docs
* updated links to lectures.quantecon.org
* tried to be more consistent with variable/value highlighting
* a lot of indenting to fix alignment of lists

@cc7768 Sorry I un-did all your unicode work but it wasn't displaying properly on my computer! Could you check the documentation now - I had to guess some of the notation. To build the docs, type `julia docs/make.jl` and open `docs/build/index.html`

* I think the structure/layout of the docs should be cleaned up - it's hard to navigate at the moment (although I don't know how to do this)
* I think this reference `scipy/scipy/optimize/Zeros/bisect.c` is outdated - might be worth linking [here](https://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.optimize.bisect.html) now
* Sometimes we say 'arguments', other times we say 'parameters'
* Not sure about the Fourier frequencies formula in estspec.jl

### DIFF
--- a/src/arma.jl
+++ b/src/arma.jl
@@ -22,7 +22,7 @@ If ``\phi`` and ``\theta`` are scalars, then the model is
 understood to be
 
 ```math
-X_t = \phi X_{t-1} + \epsilon_t + \theta \epsilon_{t-1}
+    X_t = \phi X_{t-1} + \epsilon_t + \theta \epsilon_{t-1}
 ```
 
 where ``\epsilon_t`` is a white noise process with standard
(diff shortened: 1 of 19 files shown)


jstac:
@natashawatkins Thanks!

cc7768:
@natashawatkins This is great! Thanks for adding this.

Your translation matched what I had written, but I've added some small updates to try and improve the notation now that we have access to latex :smile:

PR #170, opened 2017-08-18
Add minimum version for Optim and NLopt

### LINKED ISSUE #169: Minimum version numbers on Optim and NLopt
@Shunsuke-Hori in order to get a new version of QuantEcon registered we will need to put minimum version numbers on packages that have had breaking changes. I know this is true for Optim -- espeically with the box constrained minimization routines. 

Would you please check to see which version of their API you are using and put a lower bound in the REQUIRE file (like we do [here](https://github.com/QuantEcon/QuantEcon.jl/blob/3e690d5c2059b82f681ee284ea81e5aef3c43213/REQUIRE#L3) for Distributions.jl)?

It might be that we need to do the same for NLopt.jl, but I don't think that API has had breaking changes.

Shunsuke-Hori:
This PR should close #169 
- `Optim` needs to be v0.7.3 or more to use `Optim.Options` [here](https://github.com/QuantEcon/QuantEcon.jl/blob/master/src/markov/markov_approx.jl#L617). (See [Optim v0.7.3 release notes](https://github.com/JuliaNLSolvers/Optim.jl/blob/master/NEWS.md#optim-v073-release-notes))
- `NLopt` itself requres at least v0.3.4 to avoid [`Uint8` in v0.3.3](https://github.com/JuliaOpt/NLopt.jl/blob/v0.3.3/src/NLopt.jl#L389). ([`UInt8` in v0.3.4](https://github.com/JuliaOpt/NLopt.jl/blob/v0.3.4/src/NLopt.jl#L389))

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -5,5 +5,5 @@ LightGraphs
 Primes
 Compat 0.18.0
 StatsBase
-Optim
+Optim 0.7.3
 NLopt
--- a/REQUIRE
+++ b/REQUIRE
@@ -6,4 +6,4 @@ Primes
 Compat 0.18.0
 StatsBase
 Optim 0.7.3
-NLopt
+NLopt 0.3.4


sglyon:
Thanks for doing this.

The Optim version you've identified seems good to me.

When you say "NLopt itself requres at least v0.3.4 ..." do you mean that NLopt requires a julia version of at least 0.3.4? If so, that should be unrelated to the version of NLopt we need for this package.

Shunsuke-Hori:
> do you mean that NLopt requires a julia version of at least 0.3.4?

No, I mean NLopt 0.3.4. 

With NLopt pinned to 0.3.3, julia 0.6.0 fails to precompile.

Shunsuke-Hori:
My English was bad. I should have written "`NLopt` needs to be at least v0.3.4 for julia 0.6.0."

sglyon:
Ok great, thanks for clarifying!

PR #176, opened 2017-10-04
WIP: Add short doc on golden_method

cc7768:
This adds documentation to the `golden_method` function.

Currently only adds for version with scalar bounds -- Will add docs to vector version as well once I make sure I understand what it does.

I will ping @sglyon once this is done.

### DIFF
--- a/src/optimization.jl
+++ b/src/optimization.jl
@@ -43,6 +43,12 @@ end
 #     out_x, out_f
 # end
 
+"""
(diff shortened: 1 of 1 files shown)


sglyon:
I merged as is, but given that @cc7768 was unsure of what was returned (the maximizer and maximal function value, respectively), we should probably document that also...

cc7768:
:astonished:

Issue #118, opened 2016-07-10
DiscreteDP Performance


MaximilianJHuber:
DiscreteDP performance might be bad in the Julia version. 

For example, this Python [code](http://nbviewer.jupyter.org/github/MaximilianJHuber/DDP/blob/master/discrete-dp-ex-py.ipynb) runs 220 times faster (local 2x2.66GHZ, 4 GB machine) than this Julia [code](http://nbviewer.jupyter.org/github/MaximilianJHuber/DDP/blob/master/discrete-dp-ex.ipynb). (The task is described [here](http://nbviewer.jupyter.org/github/QuantEcon/QuantEcon.applications/blob/master/discrete_dp/discrete_dp_solutions_py.ipynb)).

Could someone please replicated this performance issue?

MaximilianJHuber:
Profiling the "solve"-function suggests nothing unusual:
![bildschirmfoto 2016-07-11 um 19 15 27](https://cloud.githubusercontent.com/assets/6653641/16759166/ed38380c-4815-11e6-8b13-6a2aa187e44e.png)

BUT: line 319 and 380 in ddp.jl have:
` vals = ddp.R + ddp.beta * ddp.Q * v`
but really should be:
` vals = ddp.R + ddp.beta * (ddp.Q * v)`.
Runtime drops by a factor of 6 and memory allocations drop from 500 MB to 2 MB of this single line!

spencerlyon2:
Great sleuthing! This makes perfect sense!

Do you mind putting that change in a pull request?

MaximilianJHuber:
I did not know what that means 30 minutes ago, but I hope I did the right thing!

MaximilianJHuber:
A new performance test with the altered matrix multiplication yielded:
constructor `0.773396 seconds (2.02 k allocations: 466.133 MB, 2.01% gc time)
`
solve `1.922052 seconds (616 allocations: 108.402 MB, 1.30% gc time)`

While Python has:
constructor `CPU times: user 103 ms, sys: 8.35 ms, total: 112 ms`
solve `CPU times: user 56.5 ms, sys: 2.05 ms, total: 58.5 ms`

So the issue remains open.

MaximilianJHuber:
Further profiling points to linear algebra to be the most expensive type of operation, for both PFI: ![untitled](https://cloud.githubusercontent.com/assets/6653641/16803570/629506f6-4907-11e6-9f58-63a843923f4a.png)
and VFI:
[...]

PR #181, opened 2017-10-19
Sl/sparse ddp q

sglyon:
When using the SA formulation of ddp, this change allows one to pass the `Q` matrix as any subtype of `AbstractMatrix`. In particular, this allows us to use a sparse matrix for `Q`. 

For the example in #118, this drops runtime of the `solve` method from 150 seconds to 0.6 seconds (ref, the python was 0.023 seconds).


Shunsuke-Hori:
@sglyon there seems to be a change in API in `Optim`. #182 should fix this.

PR #182, opened 2017-10-20
update syntax

Shunsuke-Hori:
Update syntax of `Optim`


sglyon:
Great thank you. 

Do we need to bump the optim version in the require file?

Shunsuke-Hori:
@sglyon updated.

sglyon:
Thanks!

PR #184, opened 2017-10-24
WIP: Fix utility functions to branch

cc7768:
The utility functions were using `ifelse` to evaluate in places where the utility function couldn't be evaluated -- This doesn't work because it evaluates both cases at once and it throws an error.

This is WIP and still needs some tests to check this and potentially some other cases.

For example

```
u = LogUtility()

# Should just do linear approximation around 1e-10 and extrapolating down, but throws error
u(-0.01) 
```

### DIFF
--- a/src/modeltools/utility.jl
+++ b/src/modeltools/utility.jl
@@ -23,9 +23,9 @@ end
 LogUtility() = LogUtility(1.0)
 
 (u::LogUtility)(c::Float64) =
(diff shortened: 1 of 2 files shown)


sglyon:
I'll hold on on review until after you notify me that it is no longer WIP

cc7768:
I think this is ready for now -- I don't have the extrapolating feature ready for elliptical utility yet, but it works for Log/CRRA/CFE

PR #193, opened 2017-11-25
BUG: Fix `simplex_grid`

oyamad:
Fix the bug reported in https://github.com/QuantEcon/Games.jl/commit/3586ebeb631678ed0428e3b487978cd867b8ca6d#commitcomment-25835913.


src/util.jl
```diff
@@ -223,7 +223,7 @@ function simplex_grid(m, n)
         x[m] = val - 1
         x[h] += 1
 
-        copy!(out, 3*(i-1) + 1, x, 1, m)
+        copy!(out, m*(i-1) + 1, x, 1, m)
```
cc7768: Ooops. Good catch.

Thanks for fixing this!

PR #196, opened 2017-12-17
QuantEcon.ecdf calls StatsBase.ecdf

### LINKED ISSUE #186: Drop ecdf?
Currently `ecdf` is available in StatsBase, leading to the warning

> both StatsBase and QuantEcon export "ecdf"; uses of it in module Main must be qualified

The StatsBase version is here:

https://github.com/JuliaStats/StatsBase.jl/blob/7e9d724cd795a719f340d6102c8b714ccec3cbde/src/empirical.jl#L5-L13

a-parida12:
Fixes #186

### DIFF
--- a/src/QuantEcon.jl
+++ b/src/QuantEcon.jl
@@ -10,9 +10,7 @@ import Distributions: pdf, skewness, BetaBinomial
 using DSP: TFFilter, freqz
 using Primes: primes
 using Compat: view, @compat
-
-using StatsBase
-import StatsBase:ecdf
+using StatsBase:ecdf
 
 @static if isdefined(Base, :Iterators)
     using Base.Iterators: cycle, take
@@ -147,7 +145,7 @@ include("markov/mc_tools.jl")
 include("markov/ddp.jl")
 include("markov/random_mc.jl")
 include("discrete_rv.jl")
-
+include("ecdf.jl")
 include("estspec.jl")
 include("kalman.jl")
 include("lae.jl")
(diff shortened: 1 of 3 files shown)


sglyon:
Thank you @a-parida12, this is the right direction.

However I noticed that the tests are failing. 

Here's my proposal for how to get them working again:

1. still `include("ecdf.jl")` in QuantEcon.jl
2. Delete everything that is currently in ecdf.jl
3. In ecdf.jl, write the necessary code so that existing usage of `QuantEcon.ECDF` doesn't throw and error (as it does in the tests). I think this means writing a single line like `@deprecate ECDF(x) StatsBase.ecdf(x)`
4. Re-run the tests to make sure they pass.

Are you comfortable giving that a shot?

src/QuantEcon.jl
```diff
@@ -11,21 +11,22 @@ using DSP: TFFilter, freqz
 using Primes: primes
 using Compat: view, @compat
 
+using StatsBase
+import StatsBase:ecdf
```
sglyon: Let's also replace these two lines with

```Julia
using StatsBase: ecdf
```

This will make Julia only bring that one function from the StatsBase module into scope inside QuantEcon, helping us avoid cluttering up our namespace.

sglyon:
The last thing we need to do is add `StatsBase` to the [REQUIRE](https://github.com/QuantEcon/QuantEcon.jl/blob/master/REQUIRE) file

a-parida12:
Hi @sglyon,
Thank you for all the suggestion.  I still have problems with codecov/project check which failed.

sglyon:
Looks good to me.

Thank you

PR #198, opened 2017-12-24
Add quadrature method

Shunsuke-Hori:
~~**NOT READY FOR REVIEW** The code seems to be working well basically, but let me finalize docs and potentially brush up the code~~

EDIT: **Ready for review**

This PR adds quadrature grid method for `discrete_var`, which is the last one of the three methods


src/markov/markov_approx.jl
```diff
@@ -13,6 +13,7 @@ https://lectures.quantecon.org/jl/finite_markov.html
 
 import Optim
 import NLopt
+import FastGaussQuadrature: gausshermite
```
sglyon: We should change this to `using FastGaussQuadrature: gausshermite`. Import has two uses:

1. Bring a module name into scope (as we do with Optim and NLopt above)
2. Bring a specific function from a different module into scope _so we can extend the function with our own methods_. We don't add methods to `gausshermite`, so we should just call `using` instead of `import`.

src/markov/markov_approx.jl
```diff
@@ -371,17 +375,17 @@ function discrete_var(b::Union{Real, AbstractVector},
             # Maximum entropy optimization
             if n_moments == 1 # match only 1 moment
                 temp[:, jj], _, _ = discrete_approximation(y1D[jj, :],
-                    X -> (reshape(X, 1, Nm)-cond_mean[jj, ii])/scaling_factor[jj],
+                    X -> (X'-cond_mean[jj, ii])/scaling_factor[jj],
```
sglyon: Do ay of these operations need `.` in front of them to make them broadcasting?

src/markov/markov_approx.jl
```diff
@@ -416,6 +420,27 @@ function discrete_var(b::Union{Real, AbstractVector},
...
+
+"""
+function warn_persistency(B::AbstractMatrix, ::Quadrature)
+    if any(eig(B)[1] .> 0.9)
+        warn("The quadrature method may perform poorly for highly persistent processes.")
+    end
+    return nothing
+end
+warn_persistency(B::Real, method::Quadrature) = warn_persistency(fill(B, 1, 1), method)
+warn_persistency(::Union{Real, AbstractMatrix}, ::Union{Even, Quantile}) = nothing
```
sglyon: Instead of `::Union{Even, Quantile}`, let's have the second argument be `::VAREstimationMethod` so  that is the default for any additional `VAREstimationMethod` subtypes we might create in the future.

Shunsuke-Hori:
@sglyon I removed `FastGaussQuadrature` from `REQUIRE` to reduce dependencies. I simply didn't know that QE has the same functionality. Regarding performance, the difference seems negligible.

sglyon:
Sounds great, thanks for the update!

PR #199, opened 2018-01-06
Add `next_k_array!` and `k_array_rank`.

shizejin:
Add `next_k_array!` and `k_array_rank` utilities, mimicking the [counterpart Python version](https://github.com/QuantEcon/QuantEcon.py/blob/e61702fd6622ae046d8252b38ab101454c2e74b2/quantecon/util/combinatorics.py#L12) implemented by @oyamad. It is mainly copying and pasting, and modifying a little part (let index start from 1 rather than 0).

### DIFF
--- a/src/QuantEcon.jl
+++ b/src/QuantEcon.jl
@@ -97,6 +97,7 @@ export
 
 # util
     meshgrid, gridmake, gridmake!, ckron, is_stable, num_compositions, simplex_grid, simplex_index,
+    next_k_array!, k_array_rank,
 
 # robustlq
     RBLQ,
--- a/src/util.jl
+++ b/src/util.jl
@@ -340,10 +340,10 @@ the descending sequences of the elements, following
 
 # Notes
 
-`InexactError` exception will be thrown if overflow occurs during
-the computation. It is the user's responsibility to ensure
-that the rank of the input array fits within the range of `T`;
-a sufficient condition for it is 
+`InexactError` exception will be thrown, or an incorrect value will be
+returned without warning if overflow occurs during the computation.
+It is the user's responsibility to ensure that the rank of the input
+array fits within the range of `T`; a sufficient condition for it is 
 `binomial(BigInt(a[end]), BigInt(length(a))) <= typemax(T)`.
 
 # Arguments
--- a/src/util.jl
+++ b/src/util.jl
@@ -270,3 +270,92 @@ function simplex_index(x, m, n)
     return idx
 end
 
+"""
+    next_k_array!(a)
+
+Given an array `a` of k distinct positive integers, sorted in
+ascending order, return the next k-array in the lexicographic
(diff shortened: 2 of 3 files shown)


oyamad:
Thanks @shizejin 

`k_array_rank` is not type stable (the return type changes according to the input *value*). Run it with `@code_warntype`.

One possibility is:

```jl
function k_array_rank(T::Type{<:Integer}, a::Vector{<:Integer})
    k = length(a)
    idx = one(T)
    for i = 1:k
        idx += binomial(T(a[i])-one(T), T(i))
    end
    return idx
end

k_array_rank(a::Vector{<:Integer}) = k_array_rank(Int, a)
```

```jl
n, k = 100, 50
a = collect((n-k+1):n)
k_array_rank(BigInt, a)
```

```jl
100891344545564193334812497256
```

```jl
k_array_rank(a)
```

```jl
InexactError()
```

I don't know if this is "Julian": @sglyon ?

sglyon:
That proposal is quite Julian. It would be great if there was a way to make it "just work" without throwing an error, but unfortunately I don't think there is.

As long as you don't expect the `InexactError` to arise often in practice, the proposal by @oyamad has my vote.

oyamad:
@sglyon Thanks.

@shizejin We may add a note about the possibility of an error as well as a sufficient condition for it not to occur, as in the [Python version](https://github.com/QuantEcon/QuantEcon.py/blob/e61702fd6622ae046d8252b38ab101454c2e74b2/quantecon/util/combinatorics.py#L108-L115).

shizejin:
@oyamad  Sure, I will modify this.

In this case, in the tournament game, should we set a threshold of the pair of n and k to determine whether use `k_array_rank(Int, a)` or `k_array_rank(BigInt, a)`? Or do we just use `k_array_rank(Int, a)` as the `InexactError` case happens almost impossibly?

oyamad:
@shizejin See [this](https://github.com/oyamad/QuantEcon.py/blob/d603c4985ae6cb4f335b67debf7c5de27db760bb/quantecon/game_theory/game_generators/bimatrix_generators.py#L438-L440).

The size of NumPy arrays is limited by the range of `np.intp`. Maybe Julia arrays are limited by the range of `Csize_t` (= `UInt`)?

shizejin:
@oyamad It seems that the `InexactError` of `binomial` happens when the function tries to convert the calculated result back to the type `T` of the arguments `n` and `k`. Therefore, follow the style of Python code, I think we can add these to Julia version of tournament game generator.

```julia
if binomial(BigInt(n), BigInt(k)) > typemax(typeof(n))
    error("Overflow! Maximum allowed size of $(typeof(n)) exceeded")
end
```

What do you think?

oyamad:
Are we talking about this PR, or the tournament game? About the latter let's discuss at `Games.jl`.

PR #202, opened 2018-02-01
RFC: Sl/qnwdist

sglyon:
This is a request for comments (RFC) on a new approach for constructing quadrature weights and nodes for any continuous, univariate distribution.

The algorithm, some properties, and even an example of intended usage is written down in the docstring for `qnwdist`. 

What thoughts does the team have about this approach? Can anyone think of a potentially better heuristic for choosing the nodes (besides the two I've implemented here).

Before merging I will write up some tests. My idea for testing would be to take a known function `f` and compare `dot(f.(nodes), weights)` to `quadgk(x -> f(x) pdf(d, x), support(d)...)` and see if our crude approximation gets close to the (hopefully accurate) adaptive Gauss Kronrod implementation  in quadgk for approximating `E[f(x) | x ~ d]`

cc @QuantEcon/lead-developers and @mwaugh0328 (who inspired me on this one)

### DIFF
--- a/src/QuantEcon.jl
+++ b/src/QuantEcon.jl
@@ -53,7 +53,8 @@ export
     discrete_var, Even, Quantile, Quadrature,
 
 # modeltools
-    AbstractUtility, LogUtility, CRRAUtility, CFEUtility, EllipticalUtility, derivative,
+    AbstractUtility, LogUtility, CRRAUtility, CFEUtility, EllipticalUtility,
+    derivative,
 
 # gth_solve
     gth_solve,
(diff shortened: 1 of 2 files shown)


sglyon:
Oh and please ignore the changes to `qnwgamma` -- I didn't change the actual algorithm, just the stopping heuristic

Flintro:
Hey @sglyon , 

I saw the request for comments and had a comment for the Gaussian-Legendre Quadrature approximation -- take this with a grain of salt and I apologise if I've misunderstood the code.

It seems to me that in selecting the nodes for this function you have used a simple transformation of the ith root for the ith legendre polynomials up to n. Is this correct? If so the approximation will not be optimal as _optimality_ would involve using each of the n roots of the **nth** legendre polynomial. That is at least according to [this link.](http://www.dam.brown.edu/people/alcyew/handouts/GLquad.pdf) 

I had a quick attempt at making the nodes correspond to the roots of the nth legendre polynomial and that seems more complicated than the existing node selection, so it might be a tradeoff here.

sglyon:
Hi @Flintro thanks for reaching out. It is entirely possible that you are correct about what that code is going. The code was ported from an old Matlab library with minimal modifications. Will you please make a new issue where we can track this discussion and perhaps post the code that you wrote so we can see where the differences are?

Thanks

PR #204, opened 2018-02-21
Add code for a multivariate discrete random variable

cc7768:
This code is very similar to the code that is in the QE.jl library for `DiscreteRV`.

We could simply treat multivariate discrete random variables as a univariate discrete random variable (and just make everything long vectors), but that requires working with conversions between linear and cartesian indexing. It also simplifies the interpretation of joint distributions.

This mostly mimics the code from `DiscreteRV` and adds a call to `ind2sub` so that it does the conversions for you.

### DIFF
--- a/src/discrete_rv.jl
+++ b/src/discrete_rv.jl
@@ -78,3 +78,69 @@ function Base.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
 end
 
 @deprecate draw Base.rand
+
+
+struct MVDiscreteRV{TV1<:AbstractArray,TV2<:AbstractVector,K,TI<:Integer}
+    q::TV1
+    Q::TV2
+    dims::NTuple{K,TI}
+
+    function MVDiscreteRV{TV1,TV2,K,TI}(q::TV1, Q::TV2, dims::NTuple{K,TI}) where {TV1,TV2,K,TI}
+        abs(sum(q) - 1.0) > 1e-10 && error("q should sum to 1")
+        abs(Q[end] - 1.0) > 1e-10 && error("Q[end] should be 1")
+        length(Q) != prod(dims) && error("Number of elements is inconsistent")
+
+        new{TV1,TV2,K,TI}(q, Q, dims)
+    end
+end
+
+
+function MVDiscreteRV(q::TV1) where TV1<:AbstractArray
+    Q = cumsum(vec(q))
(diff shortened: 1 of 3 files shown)


sglyon:
Just throwing this out -- would it help to replace the `q` and `Q` fields with an instance of `DiscreteRV` on `vec(q)`?

src/discrete_rv.jl
```diff
@@ -78,3 +78,69 @@ function Base.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
...
+        abs(sum(q) - 1.0) > 1e-10 && error("q should sum to 1")
+        abs(Q[end] - 1.0) > 1e-10 && error("Q[end] should be 1")
+        length(Q) != prod(dims) && error("Number of elements is inconsistent")
+
+        new{TV1,TV2,K,TI}(q, Q, dims)
+    end
+end
+
+
+function MVDiscreteRV(q::TV1) where TV1<:AbstractArray
```
sglyon: I think it makes sense to have the only inner constructor take just `q`. We wouldn't have to do the `length(Q) == prod(dims)` check 

We don't really want people to create these in any other way

src/discrete_rv.jl
```diff
@@ -78,3 +78,69 @@ function Base.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
...
+##### Arguments
+
+- `d::MVDiscreteRV`: The `DiscreteRV` type representing the distribution
+- `k::Int`
+
+##### Returns
+
+- `out::Vector{NTuple{Int}}`: `k` draws from `d`
+"""
+Base.rand(d::MVDiscreteRV{T1,T2,K,TI}, k::V) where {T1,T2,K,TI,V} =
```
sglyon: What is `k` supposed to be? I think it should have `k::Integer` instead of a free type param

src/discrete_rv.jl
```diff
@@ -78,3 +78,69 @@ function Base.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
...
+    NTuple{K,TI}[rand(d) for i in 1:k]
+
+function Base.rand!(out::AbstractArray{NTuple{K,TI}}, d::MVDiscreteRV) where {K,TI}
```
sglyon: Will you please add a tests for this method?

PR #207, opened 2018-04-29
Drop `num_actions` from DiscreteDP

a-parida12:
solves #103

### DIFF
--- a/src/markov/ddp.jl
+++ b/src/markov/ddp.jl
@@ -68,11 +68,11 @@ mutable struct DiscreteDP{T<:Real,NQ,NR,Tbeta<:Real,Tind,TQ<:AbstractArray{T,NQ}
         (beta < 0 || beta >= 1) &&  throw(ArgumentError("beta must be [0, 1)"))
 
         # verify input integrity 2
-        num_states, num_actions = size(R)
-        if size(Q) != (num_states, num_actions, num_states)
+        n, m = size(R)
+        if size(Q) != (n, m, n)
             throw(ArgumentError("shapes of R and Q must be (n,m) and (n,m,n)"))
         end
-
+	num_states=n
         # check feasibility
         R_max = s_wise_max(R)
         if any(R_max .== -Inf)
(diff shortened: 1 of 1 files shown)


sglyon:
Thanks for taking this on!

I think my preference would be to just delete the function, but maintain the naming of variables inside our other functions (e.g. still use `num_states, num_actions` instead of `m, n`).

I'd like to get some other opinions on that though... cc @oyamad @cc7768 @mmcky @jstac

a-parida12:
@sglyon I don't fully understand which particular function to be dropped then.

sglyon:
Sorry for not being clear on this one.

You did delete the proper function, but in a few instances you also changed the name of some variables (in the DiscreteDP constructor as well as a few docstrings) away from num_actions to the single letter n.

We should keep the descriptive variable names, but still delete the function as you did. I think that we want to end up with a 1-line change relative to the current master branch that just removes the function definition. We should also check `src/QuantEcon.jl` to make sure `num_actions` is not being exported.

src/markov/ddp.jl
```diff
@@ -577,7 +577,7 @@ end
 
 """
 Return the `Vector` `max_a vals(s, a)`,  where `vals` is represented as a
-`AbstractMatrix` of size `(num_states, num_actions)`.
+`AbstractMatrix` of size `(num_states, num_actions).
```
sglyon: I think here we need a closing backtick character after `num_actions)

src/markov/ddp.jl
```diff
@@ -589,7 +589,7 @@ s_wise_max!(vals::AbstractMatrix, out::AbstractVector) = (println("calling this
 
 """
 Populate `out` with  `max_a vals(s, a)`,  where `vals` is represented as a
-`AbstractMatrix` of size `(num_states, num_actions)`.
+`AbstractMatrix` of size `(num_states, num_actions).
```
sglyon: Same comment about backtick here

a-parida12: yes, my bad. I will improve the formatting.

sglyon:
Thank you

PR #213, opened 2018-08-09
Fix 0.7 errors (for using QuantEcon)

oyamad:
Enables `using QuantEcon` in v0.7 (some tests do not pass though).

### DIFF
--- a/src/discrete_rv.jl
+++ b/src/discrete_rv.jl
@@ -53,7 +53,7 @@ Make a single draw from the discrete distribution.
 
 - `out::Int`: One draw from the discrete distribution
 """
-Base.rand(d::DiscreteRV) = searchsortedfirst(d.Q, rand())
+Random.rand(d::DiscreteRV) = searchsortedfirst(d.Q, rand())
 
 """
 Make multiple draws from the discrete distribution represented by a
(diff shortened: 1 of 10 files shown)


src/discrete_rv.jl
```diff
@@ -68,13 +68,13 @@ Make multiple draws from the discrete distribution represented by a
...
-function Base.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
+function Random.rand!(out::AbstractArray{T}, d::DiscreteRV) where T<:Integer
     @inbounds for I in eachindex(out)
         out[I] = rand(d)
     end
     out
 end
 
-@deprecate draw Base.rand
+@deprecate draw Random.rand
```
sglyon: I think we can now drop this deprecation -- it has been here for a while and when we release QuantEcon.jl again it will be a new minor version, which gives us room to make breaking changes.

sglyon:
I think at this stage we should really commit to supporting julia 0.7 by not putting 0.7 in the allowed failures section of the travis file. Would you mind doing that here and we can use this branch/PR as the staging grounds for all the necessary 0.7 changes?

oyamad:
@sglyon I would like to have this PR merged now, for the following reason:

Getting `QuantEcon.jl` to work with 0.7 is highly nontrivial: there are so many things to fix. I would like to work on 0.7 compatibility issues also for `Games.jl` for example. Even `using Games` doesn't work without `using QuantEcon` to work. This PR just gets `using QuantEcon` to work, without affecting the compatibility for 0.6 (the last two "COMPAT" commits are just to reduce deprecation warnings). What do you think?

sglyon:
I'm ok with that. 

Merging now!

PR #214, opened 2018-08-10
COMPAT: Use cholesky for cholfact

oyamad:
To have `MVNSampler` work in v0.7 by avoiding `RankDeficientException` (used in `Games.jl`).


sglyon:
Thank you!

PR #215, opened 2018-08-21
WIP: Tagging a 1.0 release with same API

Nosferican:
Things still needing attention
- Iteration protocol
  - `MCIndSimulator`
  - `MCSimulator`
  - `@def_sim`
- Test_at_def_sim
- Clean up of a few deprecations (mostly finding out the scalars for `.+` and `.-`
- Verify everything
- Clean up Project and Manifest

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ addons:
     - libnlopt0
     - hdf5-tools
 julia:
-    - 0.7
+    - 1.0
 matrix:
     allow_failures:
 notifications:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,13 +8,9 @@ addons:
     - libnlopt0
     - hdf5-tools
 julia:
-    - 0.6
     - 0.7
-    - nightly
 matrix:
     allow_failures:
-        - julia: 0.7
-        - julia: nightly
 notifications:
     email: false
 #script: # use the default script setting which is equivalent to the following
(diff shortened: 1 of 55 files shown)


Nosferican:
@sglyon I understand you are pretty busy these days. If you find a bit of time, I would greatly appreciate some general guideline on `@def_sim` and its test module. I am still trying to get the iteration protocols updated, but since they call each other it was a harder to unit test... Any feedback would be great!

Nosferican:
Right now I am having issues `test/test_modeltool.jl` ln 75 (basically need to update that macro in `` `src/types.jl` including the iteration protocols)
```
@def_sim Simulation (T => Float64,) struct Observation{T<:Number}
LoadError: type LineNumberNode has no field head
```
The second issue is `mc/tools.jl` the iteration protocol for `MCIndSimulator` and `MCSimulator`. The issue is that updating one will break the other and don't know a good replacement for `Base.done(mcs::MCSimulator, state::Tuple{Int,Int}) = done(mcs.mcis, state)` now that `done` is gone in the new protocol.

sglyon:
Ahh ok. Yes that macro leveraged some of the Julia AST that has changed between 0.6 and 0.7. Will need to re-write them. 

- - -

For iterators... It sounds like you want to update just one of them. Is that right? I would think that updating both at the same time would be simpler. 

The implementation there has the MCSimulator wrap an MCIndSimulator. Once we have an implementation of the new protocol for the latter, we should get it almost for free for the former.

Do you have code that implements the new protocol for `MCIndSimulator`?

Nosferican:
@sglyon There were some issues that Git was not replacing the files from `.julia/dev/QuantEcon` to `Github/QuantEcon.jl`. Had to copy paste the files directly, but now these pass 0.7 (except for the macro AST thing we talked about).

sglyon:
Great, thank you for all the effort here! 

It will take me a bit of time to digest all the changes, but we'll get it sorted out very soon

PR #221, opened 2018-09-07
Update installation for Julia 1.0

natashawatkins:
Is this descriptive enough? Should we keep installation instructions for sub-1.0?


sglyon:
Looks great, thanks!

And no need to keep old Julia instructions because the master branch no longer supports them

PR #222, opened 2018-10-03
[WIP] fix incorrect n

Shunsuke-Hori:
test will be added soon


Shunsuke-Hori:
the error of codecov does not seem to be my fault...

PR #223, opened 2018-10-08
add Baxter King filter

Shunsuke-Hori:
test is added too

### DIFF
--- a/src/filter.jl
+++ b/src/filter.jl
@@ -85,3 +85,34 @@ function hamilton_filter(y::AbstractVector, h::Integer)
     y_trend = y - y_cycle
     return y_cycle, y_trend
 end
+
+@doc doc"""
+This function applies "Baxter King filter" to `y <:AbstractVector`.
+
+Reference: https://www.mitpressjournals.org/doi/abs/10.1162/003465399558454?casa_token=pfn7A97wi0QAAAAA:78AlfSOyiz__a6_snyH7jwydLu8uGZV-U3ZVY5Vo3dTi0r7da5uuVtUyUR87-uZBAuarQAYMiywJ76o
+
+##### Arguments
+- `y::AbstractVector` : data to be filtered
+- `wu::Real` : upper cutoff frequencies
+- `wl::Real` : lower cutoff frequencies
+- `K::Integer` : number of lead and lag of the moving average
+
+##### Returns
+- `y_cycle::Vector` : cyclical component
+- `y_trend::Vector` : trend component
+"""
+function bk_filter(y::AbstractVector, wu::Real, wl::Real, K::Integer)
+    T = length(y)
+    w1 = 2pi/wu
+    w2 = 2pi/wl
+    b = vcat((w2-w1)/pi, [(sin(w2*i)-sin(w1*i))/(i*pi) for i=1:K])
+    theta = (b[1] + 2*sum(b[2:end]))/(2K+1)
+    B = b .- theta
+    y_cycle = Vector{Union{Missing, Float64}}(undef, T)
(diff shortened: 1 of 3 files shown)


sglyon:
Hi @Shunsuke-Hori thanks for submitting!

I googled for some references on the BK filter equations and found conflicting evidence. 

Two references are

- http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.197.8691&rep=rep1&type=pdf
- http://help.prognoz.com/en/mergedProjects/Lib/02_time_series_analysis/uimodelling_baxterkingfilter.htm

A few differences:

- the html page only divides `b[2:end]` by (2K+1), but keeps `b[1]` without dividing
- The pdf has a negative sign in front of theta, while the html page keeps it positive. 


your code seems to more closely follow the pdf, but I'm not sure.

What reference did you use? How do we know which is correct?

Shunsuke-Hori:
@sglyon Thank you for the review.

I first looked the html but since the output was not same as python's, I switched to the pdf. 

I checked the original paper and it seems the pdf is consistent with the paper (although, honestly, I mindlessly trusted the result of [`bk_filter' of python](http://www.statsmodels.org/stable/generated/statsmodels.tsa.filters.bk_filter.bkfilter.html#statsmodels.tsa.filters.bk_filter.bkfilter) when I wrote the code)

andi3141:
Hi @Shunsuke-Hori 
I had a look at the code. It seems perfectly in line with 
http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.197.8691&rep=rep1&type=pdf
The second reference
http://help.prognoz.com/en/mergedProjects/Lib/02_time_series_analysis/uimodelling_baxterkingfilter.htm
seems not to be online any more. Further the original reference
https://www.mitpressjournals.org/doi/abs/10.1162/003465399558454?casa_token=pfn7A97wi0QAAAAA:78AlfSOyiz__a6_snyH7jwydLu8uGZV-U3ZVY5Vo3dTi0r7da5uuVtUyUR87-uZBAuarQAYMiywJ76o
is not publicy available (MIT press). I try to get my hands on that at some point.
The implementation gives the same result as the python implementation in 
https://www.statsmodels.org/stable/generated/statsmodels.tsa.filters.bk_filter.bkfilter.html#statsmodels.tsa.filters.bk_filter.bkfilter

So I think, it is fine.

src/filter.jl
```diff
@@ -85,3 +85,34 @@ function hamilton_filter(y::AbstractVector, h::Integer)
...
+@doc doc"""
+This function applies "Baxter King filter" to `y <:AbstractVector`.
+
+Reference: https://www.mitpressjournals.org/doi/abs/10.1162/003465399558454?casa_token=pfn7A97wi0QAAAAA:78AlfSOyiz__a6_snyH7jwydLu8uGZV-U3ZVY5Vo3dTi0r7da5uuVtUyUR87-uZBAuarQAYMiywJ76o
+
+##### Arguments
+- `y::AbstractVector` : data to be filtered
+- `wu::Real` : upper cutoff frequencies
+- `wl::Real` : lower cutoff frequencies
+- `K::Integer` : number of lead and lag of the moving average
```
jstac, line 98: Should be `leads and lags`

jstac:
Many thanks @andi3141 for your review.  This looks good to me too (aside from very minor comment above).

Shunsuke-Hori:
Thank you for reviewing, @andi3141 and @jstac 

I fixed the grammar.

jstac:
Looks good to me.  Over to you @sglyon

PR #225, opened 2018-10-16
Add inv and derivativeinv to some utility functions

cc7768:
Adds methods for `u^{-1}` and `du^{-1}` for utility functions where it can be computed by hand.

I have some tests that make sure `u(u^{-1})` returns what we expect, but it wouldn't hurt to have someone glance at the equations to make sure I didn't do anything stupid.

### DIFF
--- a/src/QuantEcon.jl
+++ b/src/QuantEcon.jl
@@ -1,6 +1,6 @@
 module QuantEcon
 
-import Base: show, isapprox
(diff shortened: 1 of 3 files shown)


sglyon:
General comment -- I think we should consider edge cases in both the implementation and the testing.

For example, what happens at `derivativeinv(u::CRRAUtility, x)` at x=0?

Issue #230, opened 2019-01-16
Error in qnwnorm


IljaK91:
https://github.com/QuantEcon/QuantEcon.jl/blob/ae0046f4a1725ce8a35719f47181a4256240c8e6/src/quad.jl#L594

As I understand, these two lines should yield the same:

```
qnwnorm(10,[1, 1],[2, 2]) 
qnwnorm([10,10],[1., 1.],[2., 2.])
```

However, the first line yields an error with
```
MethodError: no method matching diagm(::Array{Float64,1})
Closest candidates are:
  diagm(!Matched::Pair{#s171,#s170} where #s170<:(StaticArrays.StaticArray{Tuple{N},T,1} where T where N) where #s171<:Val...) at C:\Users\Ilja\.julia\packages\StaticArrays\mcf7t\src\linalg.jl:177
  diagm(!Matched::Pair{#s576,#s575} where #s575<:(AbstractArray{T,1} where T) where #s576<:Integer...) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\stdlib\v1.0\LinearAlgebra\src\dense.jl:313
  diagm(!Matched::StaticArrays.StaticArray{Tuple{N},T,1} where T where N) at deprecated.jl:53
  ...
qnwnorm(::Int64, ::Array{Int64,1}, ::Array{Int64,1}) at quad.jl:594
top-level scope at none:0
```

It seems to me that an easy fix would be to use `Matrix(Diagonal())` instead of `diagm()`

IljaK91:
I created a pull request.

PR #231, opened 2019-01-16
Changing diagm to Matrix(Diagonal()) in qnwnorm()

IljaK91:
Changed diagm() to Matrix(Diagonal())


sglyon:
Thank you for the fix!

closes #230

PR #232, opened 2019-01-30
Kalman Filter: Fix problem when y is a scalar but x is a vector

IljaK91:
There need to be some dots added. These are the two that I encountered. It might be necessary somwhere else as well.

Example:
```
using QuantEcon

A = [0. 0.; 0. 1]
G = [1. 0.5]
R = 0.
Q = [1 0.; 0. 1]

k = Kalman(A,G,Q,R)
stationary_values(k)
prior_to_filtered!(k, 1.)
```


sglyon:
Great, thank you!

PR #235, opened 2019-02-09
Made a kalman_filter method similar to kalman_smooth.

murphyk:
Added kalman_sample to generate from model.
Added example of 2d Kalman filtering and smoothing.

### DIFF
--- a/examples/kalman_tracking_2d_example.jl
+++ b/examples/kalman_tracking_2d_example.jl
@@ -1,0 +1,66 @@
+using  LinearAlgebra,  Distributions
+
+function make_params()
+    F = [1 0 1 0; 0 1 0 1; 0 0 1 0; 0 0 0 1.0];
+    H = [1.0 0 0 0; 0 1 0 0];
+    nobs, nhidden = size(H)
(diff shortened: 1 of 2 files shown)


sglyon:
Thank you for your contribution!

Before we merge pull requests that add new library code, we ask that contributors write simple tests to ensure that their code functions as expected.

Would you be willing to do that?

I see that you provided an example (thank you!). Perhaps the example code could be re-used for the tests.

murphyk:
I have added some extra tests to test_kalman.jl. They pass locally on my laptop but Travis complains, I don't know why. (I am new to Julia and open source development...) I have temporarily commented out the new tests. Note that the call to `smooth` actually calls my new `filter` function, so is an indirect test of correctness :)  But a direct test would be better. Would you mind telling me what I'm doing wrong?

sglyon:
Hey @murphyk from what I can tell the error on Travis has to do with a no method error when trying to call a `Kalman` instance like a function: https://travis-ci.org/QuantEcon/QuantEcon.jl/builds/491097872#L816

Does that sound familiar to you?

PR #240, opened 2019-03-06
DOC: Update

### LINKED ISSUE #239: Documenter not working
See https://travis-ci.org/QuantEcon/QuantEcon.jl/jobs/501549954#L986

Cf. QuantEcon/Games.jl#99

shizejin:
closes #239.

Update keyword value for `Documenter.makedocs` and set `prettyurls` off as in QuantEcon/Games.jl#106.

It seems that we cannot break a math expression in `@doc` block if we use "$$", as I modified in `util.jl`. Otherwise it will trigger documentation rendering failure.

### DIFF
--- a/src/util.jl
+++ b/src/util.jl
@@ -236,8 +236,8 @@ end
 
 @doc doc"""
 Return the index of the point x in the lexicographic order of the
-integer points of the (m-1)-dimensional simplex $\{x \mid x_0
-+ \cdots + x_{m-1} = n\}$.
+integer points of the (m-1)-dimensional simplex $\{x \mid x_0 + 
+\cdots + x_{m-1} = n\}$.
 
 ##### Arguments
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,14 @@ matrix:
 notifications:
     email: false
 #script: # use the default script setting which is equivalent to the following
+jobs:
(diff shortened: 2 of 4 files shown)


src/util.jl
```diff
@@ -236,8 +236,8 @@ end
 
 @doc doc"""
 Return the index of the point x in the lexicographic order of the
-integer points of the (m-1)-dimensional simplex $\{x \mid x_0
-+ \cdots + x_{m-1} = n\}$.
+integer points of the (m-1)-dimensional simplex 
+$\{x \mid x_0 + \cdots + x_{m-1} = n\}$.
```
oyamad: Isn't it that a symbol `+` should not be placed at the beginning of a line?

Would the following work?

```
integer points of the (m-1)-dimensional simplex $\{x \mid x_0 +
\cdots + x_{m-1} = n\}$.
```

shizejin: You are correct. It worked. Thank you for pointing it out!

sglyon:
Thank you for working on this @shizejin 

Please let me know when it is ready for review

shizejin:
@sglyon I think it is now ready :)

sglyon:
Ok, nothing controversial here. Let's give it a shot!

Thank you

PR #245, opened 2019-04-29
Fix bug in random_probvec

oyamad:
Fix the bug:

```jl
QuantEcon.random_probvec(5)
```

```
UndefVarError: rng not defined
```


sglyon:
Thanks!

PR #247, opened 2019-08-20
fixed DSP.TFFilter deprecation

guihigashi:
also updated dependencies because `test QuantEcon` was giving errors

### DIFF
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -382,7 +382,7 @@ uuid = "4c63d2b9-4356-54db-8cca-17b64c39e42c"
 version = "0.8.0"
 
 [[SuiteSparse]]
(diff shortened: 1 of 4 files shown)


sglyon:
Could you also update the Project.toml file to make sure the dependency for the version of DSP containing PolynomialRatio is required?

guihigashi:
I checked that DSP v0.5.2 is not compatible with julia 1.0, so DSP v0.6.0 is the minimum.

I'm new to git, I thought it was best to amend last commit. If that's not alright I guess I can make a new one.

sglyon:
Thank you for adding. I noticed you closed the pull request. Was this intentional?

PR #250, opened 2019-10-08
Add filer and smoothing for regime switching model 

Shunsuke-Hori:
Filtering and smoothing features for regime switching model are added. The tests are added too.


src/filter.jl
```diff
@@ -85,3 +85,206 @@ function hamilton_filter(y::AbstractVector, h::Integer)
...
+               parameter = ParaType([0, 0.5], [3, 4])
+               ```
+- `P::AbstractMatrix`: Transition matrix of state. It must be square.
+- `M::Integer`: Number of states. It must be same as the number of rows and columns of `P`. If `M` is not passed, 
+                number of rows of `P` is set as `M`.
+"""
+mutable struct RegimeSwitchingModel{TD <: AbstractArray, TP <: AbstractMatrix, TPS <: AbstractMatrix}
+    g::Function
+    y::TD
+    parameter
```
cc7768: Is there a reason that we don't want this typed? By having a mutable structure and not having this argument typed, we can change it to anything we want -- For example, I can set `rsm.parameter = 1` and it won't complain. Is there a case where this changes types?

src/filter.jl
```diff
@@ -85,3 +85,206 @@ function hamilton_filter(y::AbstractVector, h::Integer)
...
+"""
+Apply the filter developed by Hamilton (1989, Econometrica) to a regime switching model.
+
+##### Arguments
+- `rsm::RegimeSwitchingModel`: `RegimeSwitchingModel`
+- `prob_update_pre::AbstractArray`: Probability distribution of state at period 0 conditional on the information up to
+                                   period 0.
+
+##### Returns
+- `p`: likelihood at each period
```
cc7768: It wouldn't hurt to emphasize that the returns are each of size `T x M` where `T` is the number of observations and `M` is the number of regimes.

src/filter.jl
```diff
@@ -85,3 +85,206 @@ function hamilton_filter(y::AbstractVector, h::Integer)
...
+- `p_s_forecast`: Probability of period `t` data conditional on the information up to period `t-1`.
+"""
+function filter!(rsm::RegimeSwitchingModel, 
+                 p_s_update::AbstractMatrix, p_s_forecast::AbstractMatrix;
+                 prob_update_pre::AbstractArray = stationary_distributions(MarkovChain(rsm.P))[1])
+    g, y = rsm.g, rsm.y
+    T = size(y, 1)
+    p = Vector{Float64}(undef, T)
+    logL = 0
+    for t in 1:T
```
cc7768: I like how the code is structured that allows you to write such a simple (and beautiful!) for-loop to do the core of the computation.

src/filter.jl
```diff
@@ -85,3 +85,206 @@ function hamilton_filter(y::AbstractVector, h::Integer)
...
+    rsm.prob_smoothed[T, :] = p_s_update[T, :]
+    for t = T-1:-1:1
+        for s in 1:M
+            for sp in 1:M
+                p_s_joint_smoothed[t, s, sp] = rsm.prob_smoothed[t+1, sp] * p_s_update[t, s] * rsm.P[s, sp]/p_s_forecast[t+1, sp]
+            end
+            rsm.prob_smoothed[t, s] = sum(p_s_joint_smoothed[t, s, :])
+        end
+    end
+    verbose ? (return p_s_joint_smoothed) : (return nothing)
```
cc7768: Does this if statement in return cause confusion to the compiler? This could prevent it from inferring the type of the output of this function. Is there a reason to not just return `p_s_joint_smoothed`?

cc7768: I don't think it does create a problem -- It identifies it as a union of nothing and the array type:

```
julia> @code_warntype test(rsm, true)
Variables
  #self#::Core.Compiler.Const(test, false)
  rsm::Main.QuantEcon.RegimeSwitchingModel{Array{Float64,2},Array{Float64,2},Array{Float64,2}}
  verbose::Bool

Body::Union{Nothing, Array{Float64,1}}
1 ─ %1 = Main.randn(10)::Array{Float64,1}
│   %2 = (:verbose,)::Core.Compiler.Const((:verbose,), false)
│   %3 = Core.apply_type(Core.NamedTuple, %2)::Core.Compiler.Const(NamedTuple{(:verbose,),T} where T<:Tuple, false)
│   %4 = Core.tuple(verbose)::Tuple{Bool}
│   %5 = (%3)(%4)::NamedTuple{(:verbose,),Tuple{Bool}}
│   %6 = Main.QuantEcon.smooth!::Core.Compiler.Const(Main.QuantEcon.smooth!, false)
│   %7 = Core.kwfunc(%6)::Core.Compiler.Const(getfield(Main.QuantEcon, Symbol("#kw##smooth!"))(), false)
│   %8 = Main.QuantEcon.smooth!::Core.Compiler.Const(Main.QuantEcon.smooth!, false)
│   %9 = (%7)(%5, %8, rsm, %1)::Union{Nothing, Array{Float64,1}}
└──      return %9

julia> @code_warntype test(rsm, false)
Variables
  #self#::Core.Compiler.Const(test, false)
  rsm::Main.QuantEcon.RegimeSwitchingModel{Array{Float64,2},Array{Float64,2},Array{Float64,2}}
  verbose::Bool

Body::Union{Nothing, Array{Float64,1}}
1 ─ %1 = Main.randn(10)::Array{Float64,1}
│   %2 = (:verbose,)::Core.Compiler.Const((:verbose,), false)
│   %3 = Core.apply_type(Core.NamedTuple, %2)::Core.Compiler.Const(NamedTuple{(:verbose,),T} where T<:Tuple, false)
│   %4 = Core.tuple(verbose)::Tuple{Bool}
│   %5 = (%3)(%4)::NamedTuple{(:verbose,),Tuple{Bool}}
│   %6 = Main.QuantEcon.smooth!::Core.Compiler.Const(Main.QuantEcon.smooth!, false)
│   %7 = Core.kwfunc(%6)::Core.Compiler.Const(getfield(Main.QuantEcon, Symbol("#kw##smooth!"))(), false)
│   %8 = Main.QuantEcon.smooth!::Core.Compiler.Const(Main.QuantEcon.smooth!, false)
│   %9 = (%7)(%5, %8, rsm, %1)::Union{Nothing, Array{Float64,1}}
└──      return %9
```

It does add "complexity" relative to `return p_s_joint_smoothed` though, so is there a reason not to return it always?

src/filter.jl
```diff
@@ -85,3 +85,206 @@ function hamilton_filter(y::AbstractVector, h::Integer)
     y_trend = y - y_cycle
     return y_cycle, y_trend
 end
+
+"""
+- `g::Function`: Parameterized `Function` that recieves three arguments
```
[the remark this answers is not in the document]
Shunsuke-Hori: @cc7768 It totally slipped my mind. I think it is 80.

src/filter.jl
Shunsuke-Hori: > Is there a reason that we don't want this typed?

Kind of, I thought some may want to use user-defined `struct` like [this](https://github.com/QuantEcon/QuantEcon.jl/pull/250/files/6fce054b38ad61c04ac13229a61f6a0707740a88#diff-e5f041a6f2fa9d7548572012875b7606R104) rather than `NamedTuple`. But I also think it should be typed. Maybe I can force it to be `NamedTuple`.

src/filter.jl
Shunsuke-Hori: That's fair. I'll do so.

src/filter.jl
Shunsuke-Hori: >  is there a reason not to return it always?

No, when I realized `p_s_joint_smoothed` should be returned, I left `return nothing` case so that I didn't need to make change to existing parts, but that was a silly idea.  Now I agree with you, it's just confusing. I'll fix it.

Shunsuke-Hori:
@cc7768 Thank you for your review!! Your comments improve the codes a lot. 

The test is based on python statsmodels, which I assumed correct. (https://www.statsmodels.org/dev/examples/notebooks/generated/markov_autoregression.html) 

It seems I was more stupid than I thought. I'll update the codes soon.

src/filter.jl
cc7768: Perfect -- This is a use case Julia should handle very cleanly! You can add a type parameter to `RegimeSwitchingModel` and then just refer to that type parameter when typing the argument. I'd make suggestion, but I can only figure out how to suggest one line change, but here's what it would look like:

```julia
mutable struct RegimeSwitchingModel{TD <: AbstractArray, TP <: AbstractMatrix, TPS <: AbstractMatrix, TPRM}
g::Function
y::TD
parameter::TPRM
...
```

This will allow the functions to specialize to the user's type (there will be a new type of  `RegimeSwitchingModel` for every type that gets put in as a parameter) and still prevent people from doing "weird" changes to the `parameter` data.

cc7768:
Excellent -- That sounds like a great place to find a test for the code :)

Many thanks for putting this together and for addressing the "nitpicks" that I have.

src/filter.jl
Shunsuke-Hori: Oh I see, that makes sense and is way better. Thank you!!

Shunsuke-Hori:
I updated the code. Could anyone do another round of review?

PR #264, opened 2020-07-02
qnwbeta: lgamma updated to logabsgamma  

Jovansam:
lgamma from "SpecialFunctions" is depracated. Updated to (logabsgamma(x))[1]


sglyon:
Thank you

PR #267, opened 2021-09-05
allowed `tauchen` to take non-integer `n_std` positional argument

amrods:
A long time ago I asked whether `tauchen` should limit `n_std` to being an `Integer`:
https://discourse.quantecon.org/t/integer-n-std-in-tauchen/541

I don't think there should be such restriction. This PR relaxes `n_std` to be a subtype of `Real`.

### DIFF
--- a/src/markov/markov_approx.jl
+++ b/src/markov/markov_approx.jl
@@ -35,7 +35,7 @@ where ``\epsilon_t \sim N (0, \sigma^2)``
 - `ρ::Real` : Persistence parameter in AR(1) process
 - `σ::Real` : Standard deviation of random component of AR(1) process
 - `μ::Real(0.0)` : Mean of AR(1) process
-- `n_std::Integer(3)` : The number of standard deviations to each side the process
+- `n_std::Real(3)` : The number of standard deviations to each side the process
   should span
 
 ##### Returns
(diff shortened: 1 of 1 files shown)


jstac:
Thanks @amrods , this seems reasonable to me.  I'll leave it to the maintainers to have their say.

amrods:
aside, it seems that the julia side of the project has been dormant. is that the case?

jstac:
It has for a while, @amrods , but there's been a heap of work behind the scenes to bring it up to date.

@mmcky and @jlperla are running the show but several RAs are also hard at work.  The updated lecture site should go live in a week or so.

jlperla:
Yes @amrods  Lots of progress.  The reason things seem dormant  is because of the move towards the new build system based on https://jupyterbook.org/intro.html  as well as a large number of changes in the notebooks to move to VS Code from Atom.

See the source in https://github.com/QuantEcon/lecture-julia.myst

jlperla:
@amrods  For this particular change:  Up to you guys as I am not a maintainerit, but the best approach is to just drop all of those type declarations entirely.  `tauchen` is not using multiple-dispatch, so it is always better to leave function arguments without a type.

You just never know what types someone will push through it.  Case in point: if there were no type declarations in the first place, then this PR wouldn't even have been necessary as everything would have just worked.

amrods:
Agreed, best not to use type declarations whenever possible. I was just following the style already present.

PR #271, opened 2022-05-13
Replace LightGraphs by Graphs, fix deprecations, update compat

greimel:
* LightGraphs has been renamed to Graphs (without change in functionality)
* there were two deprecation warnings in the tests that I fixed
* updated SpecialFunctions compat

I had to remove on test that relies on simulation/estimation. Since Travis CI has not run in a while I am not sure if this is an actual regression.

PR #272, opened 2022-05-16
replace Travis CI by Github Actions

greimel:
This is factored out of my other PR #271. I've tested this PR on my fork (see https://github.com/greimel/QuantEcon.jl/runs/6448797672?check_suite_focus=true)

@sglyon, please enable workflows for this PR (this is necessary for all first time contributors)

You will see that this one test is passing on 1.6 but failing on 1.7

The test depends on generated random numbers and these are unstable across Julia versions.

* One way to fix this would be using https://github.com/JuliaRandom/StableRNGs.jl but I have no bandwidth to implement this. 
* Another option would be to allow a failure on this test, open an issue and fix this in the future.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix: 
+        version:
+          - '1.6' # LTS
+          - '1' # expands to latest 1.x release
```
greimel, line 20:
```suggestion
```

@sglyon would you be fine with only testing on 1.6 for now (all tests pass there!).

I'd appreciate a response here (and #271), QuantEcon is holding back quite a few packages that I'd like to use for assignments.

greimel: removed this line to ensure that tests will pass

oyamad:
@greimel Thank you for the PR!
I think the package should be tested also against the current version (= 1.7) (and nightly also), even if some tests fail.

greimel:
Thanks for the feedback @oyamad!

I've updated the PR. Could you please approve the CI run again?

greimel:
Thanks for approving the CI run! As you can see, there is one failing test on latest and nightly

image

I think this should be fixed in a separate PR. Would you mind merging this?

oyamad:
@greimel Thanks! Merging.

Issue #273, opened 2022-07-01
Fix test failure in test_markov_approx.jl


oyamad:
https://github.com/QuantEcon/QuantEcon.jl/runs/7090838653?check_suite_focus=true#step:6:211

See #272.

oyamad:
Done by #274.

PR #274, opened 2022-07-01
Fix broken test on julia 1.7+

greimel:
This is a quick and dirty fix for #273. Ideally one would use stable RNGs in the tests. But that is beyond my capacities.


oyamad:
@greimel Thanks!

Would you mind opening an issue explaining about an ideal fix that uses "stable RNGs in the tests"? Thank you!

PR #277, opened 2022-07-23
fix default value and change variable name of discrete_approximation

Shunsuke-Hori:
- Default value wasn't properly set. 
    - `N` was not defined. 
    - `Tbar` was not defined (though `TBar` was defined).
- I set initial value more explicitly by using `length(D)`
- I renamed `TBar` into `Tbar`. I think it's better as a variable that represents $\overline{T}$.


Shunsuke-Hori:
It's ready for a review. Would anyone be able to review this PR?

PR #392, opened 2026-07-10
FIX: Fix known bugs in MarkovChain (markov/mc_tools.jl)

oyamad:
This PR fixes several bugs in `markov/mc_tools.jl` found in a code review, following up on #385/#391.

## Fixes

- `gth_solve`/`gth_solve!` silently returned an incorrect result (NaN-contaminated) for non-square input, since only `size(A, 1)` was read and the elimination runs under `@inbounds`. A `DimensionMismatch` is now thrown.
- `gth_solve!` allocated temporaries in the elimination loop (row-slice copy in the pivot-scale sum and an out-of-place column scaling) and in the final normalization. These are now views/in-place: at n=200, allocations drop from 1201 (842 KiB) to 5 (322 KiB, essentially the input copy in `gth_solve`), with a ~10% speedup at small sizes (benchmark `SUITE["mc_tools"]["gth_solve"]`, #391).
- `check_stochastic_matrix` used a fixed absolute tolerance of 5e-15 on row sums, which rejects valid large matrices: both the rounding of the entries (e.g. from row normalization) and the accumulation in the recomputed sum grow linearly with n, so e.g. a row-normalized random 1000x1000 matrix typically fails the check. The tolerance is now `max(5e-15, n * eps(T))` for `T<:AbstractFloat` (with `eps(Float64)` otherwise); the floor keeps the current behavior for small matrices. This also makes `Float32` transition matrices usable, which the fixed tolerance rejected essentially always.
- `test_mc_tools.jl` defined a `Base.isapprox` overload for `Vector{Vector{<:Real}}` — type piracy that in fact never dispatched (by type invariance, `Vector{Vector{Float64}} <: Vector{Vector{<:Real}}` does not hold), and whose body compared the outer arguments, so it would have recursed infinitely if it ever ran. It is removed; the tests exercise the generic `isapprox` fallback as they always effectively did.

## Tests

New test sets cover the non-square `DimensionMismatch` (for the `Float64`, `Int`, and in-place entry points), row-sum errors within and beyond the size-dependent tolerance, and construction from a row-normalized `Float32` matrix.

### DIFF
--- a/src/markov/mc_tools.jl
+++ b/src/markov/mc_tools.jl
@@ -14,7 +14,13 @@ https://lectures.quantecon.org/jl/finite_markov.html
 import Graphs: DiGraph, period, attracting_components,
                strongly_connected_components, is_strongly_connected
 
-@inline check_stochastic_matrix(P) = maximum(abs, sum(P, dims = 2) .- 1) < 5e-15 ? true : false
+# Row sums can deviate from 1 by rounding of the entries and accumulation
+# in the sum, both growing linearly with the matrix size
+@inline function check_stochastic_matrix(P)
+    T = eltype(P)
+    atol = max(5e-15, size(P, 1) * (T <: AbstractFloat ? eps(T) : eps()))
+    return maximum(abs, sum(P, dims = 2) .- 1) <= atol
+end
 
 """
     MarkovChain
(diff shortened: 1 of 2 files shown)


oyamad:
A review by ChatGPT caught three genuine flaws in the tolerance introduced here, fixed in 16f3575. (1) `n * eps(T)` was unbounded and reached exactly 1 for Float16 at n=1024, accepting an all-zero matrix; the tolerance is now capped at `sqrt(eps)` so it can never approach the scale of the entries. (2) For sparse matrices the accumulation length is now the maximum number of stored entries per row rather than n, so single-entry rows 1% short of 1 are rejected at any size. (3) Half/single-precision row sums are now accumulated in `Float64` (`P * ones(Float64, n)`), so the check itself adds no low-precision rounding, and abstract element types fall back to `eps(Float64)` instead of raising a `MethodError`. Regression tests added for each rejected/accepted example; the existing normalized-Float32 acceptance test is retained.

Issue #125, opened 2016-07-13
MarkovChain: Respect sparsity


oyamad:
In the current implementation of `MarkovChain`, if the input is a sparse matrix, it is converted to dense in `simulation` methods. The following is a possible approach to keeping sparsity of the input matrix.

```jl
type SparseMarkovChain{T,Ti<:Integer,TV<:AbstractVector}
    P::SparseMatrixCSC{T,Ti}
    n::Int
    rowptr::Vector{Ti}
    colval::Vector{Ti}
    cdfs::Vector{T}
    state_values::TV
end
```

In the constructor of `SparseMarkovChain`, create `rowptr`, `colval`, and `nzval` from the input `SparseMatrixCSC` by using `csr2csc` from [SparseMatrices.jl](https://github.com/JuliaSparse/SparseMatrices.jl), and generate `cdfs` as in the [Python version](https://github.com/QuantEcon/QuantEcon.py/blob/master/quantecon/markov/core.py#L437-L448).

(@spencerlyon2 What is the status of SparseMatrices.jl?)

Extend the `DiscreteRV`:

```jl
type DiscreteRV{T1<:AbstractVector,T2<:AbstractVector,TV<:AbstractVector}
    pdf::Nullable{T1}
    cdf::T2
    values::TV
end

draw(d::DiscreteRV) = d.values[searchsortedfirst(d.Q, rand())]
```

In `simulate` methods, use `DiscreteRV` as follows:

```jl
function simulate!(mc::SparseMarkovChain{T,Ti}, X::Matrix{Ti})
    P_dist = [
        DiscreteRV(
            Nullable(),
            sub(mc.cdfs, mc.rowptr[i]:mc.rowptr[i+1]-1),
            sub(mc.colval, mc.rowptr[i]:mc.rowptr[i+1]-1),
        ) for i in 1:mc.n
    ]
    ...
end
```

For computation of stationary distributions, we can implement an iterative method such as Successive Over-Relaxation (SOR), for which CSC is more appropriate.

spencerlyon2:
@oyamad interesting proposal. I would really like to be able to preserve the sparsity of the chain when simulating. 

I've done a bit more work with SparseMatrices.jl that I don't believe I have published. I would love to be able to flesh that project out a bit more, but I simply haven't had the time. 

I think the proposed extension to DiscreteRV sounds good. I'd move the `pdf::Nullable` field to the end and fill it with a default value like this:

```julia
[...]

PR #395, opened 2026-07-10
PERF: Speed up MarkovChain simulation; add dedicated sparse path

oyamad:
This PR addresses the performance findings of the `mc_tools.jl` review (following #391/#392), including the simulation part of #125.

## Changes

- The transition CDFs used by the simulation routines are now computed once and cached on the `MarkovChain` instance (lazily, on the first simulation call). Previously they were rebuilt on every `simulate` call, through a lazy adjoint whose columns are stride-n views — the pre-0.7 "transpose then view" optimization that `Matrix(mc.p)'` had silently voided. The cache field is typed `Union{Nothing,TS}` with `TS` a new type parameter computed from the matrix type, so the sampler access and the simulation entry points are type stable (tested with `@inferred`). Assigning a new matrix to `mc.p` invalidates the cache via a `Base.setproperty!` override; modifying the matrix in place does not, which is documented in the `MarkovChain` docstring.
- Sparse transition matrices get a dedicated sampler that draws from per-row nonzero-entry CDFs (stored as the materialized transpose), instead of being converted to dense.
- Both samplers guard the CDF search: a random draw exceeding the final CDF value (possible when a row sum falls slightly short of 1 by rounding) falls back to the last state with positive transition probability, as in QuantEcon/QuantEcon.py#860. The previous `DiscreteRV`-based path could emit the out-of-range index n+1 in that event.
- `recurrent_classes` returns the single recurrent class directly for strictly positive dense matrices (which are irreducible) instead of building the graph; this removes ~1 ms of the 1.81 ms of `stationary_distributions` at n=200.
- New tests: cache invalidation on `mc.p` assignment, dense/sparse same-seed path equality, a `Bool`-eltype chain (the dense CDFs take the accumulation type of `cumsum`), the overflow fallback for both samplers, and `@inferred` on the constructor and the simulation path. The dense sampler draws the same paths as the previous implementation for a given random stream, so existing seeded results are unchanged.

## Benchmarks

`SUITE["mc_tools"]`, minimum times, against the #391 baseline; Python = warm-cache QuantEcon.py on the same machine:

| Benchmark | Baseline | This PR | Python (warm) |
|:--|--:|--:|--:|
| `simulate/dense_n1000_ts100` | 928 μs | 3.6 μs | 10.4 μs |
| `simulate/sparse_n1000_k4_ts10000` | 1.83 ms | 138 μs | 191 μs |
| `stationary_distributions/dense_n200` | 1.81 ms | 844 μs | 925 μs |
| `simulate/dense_n100_ts10000` | 307 μs | 263 μs | 296 μs |
| `simulate!/dense_n100_10000x10` | 2.97 ms | 2.62 ms | 2.91 ms |

A second commit adds the benchmark cases deferred from the #391 review: `_cold` simulate variants (fresh chain per sample, timing the CDF-cache construction — 846 μs / 150 μs), a reducible `stationary_distributions` case exercising the graph path that positive matrices now bypass (538 μs), and README documentation of the warm/cold semantics.

### DIFF
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -71,17 +71,10 @@ irreducible), each generated with its own fixed-seed RNG:
 | `stationary_distributions/dense_n200`, `stationary_distributions/sparse_n300_k4` | Recurrent class detection + GTH solve |
 | `stationary_distributions/dense_n200_reducible` | Two recurrent classes; exercises the graph path, which strictly positive matrices bypass |
 | `simulate/dense_n100_ts10000` | Long path: per-step sampling dominates |
-| `simulate/dense_n1000_ts100` | Short path, many states |
+| `simulate/dense_n1000_ts100` | Short path, many states: per-call setup (transition-CDF construction) dominates |
 | `simulate/sparse_n1000_k4_ts10000` | Sparse transition matrix (dedicated sparse sampler) |
-| `simulate/{dense_n1000_ts100,sparse_n1000_k4_ts10000}_cold` | As above, on a freshly constructed chain: the timed call includes building the CDF cache |
 | `simulate!/dense_n100_10000x10` | 10 paths into a preallocated matrix |
 | `simulate_indices/dense_n100_ts10000` | Long path, index-valued output |
-
-The `MarkovChain` objects of the `simulate` cases are shared across
-samples, so those cases measure the warm-cache steady state: the
-transition CDFs are computed and cached on the instance on first use.
-The `_cold` cases construct a fresh chain in `setup` to time the
-first-use cost.
 
 ## Running the suite standalone
 
--- a/benchmark/README.md
+++ b/benchmark/README.md
@@ -69,11 +69,19 @@ irreducible), each generated with its own fixed-seed RNG:
 | `gth_solve/n{50,200,1000}` | GTH solver on a dense stochastic matrix |
 | `constructor/dense_n100`, `constructor/sparse_n1000_k4` | `MarkovChain` construction (input verification) |
 | `stationary_distributions/dense_n200`, `stationary_distributions/sparse_n300_k4` | Recurrent class detection + GTH solve |
+| `stationary_distributions/dense_n200_reducible` | Two recurrent classes; exercises the graph path, which strictly positive matrices bypass |
 | `simulate/dense_n100_ts10000` | Long path: per-step sampling dominates |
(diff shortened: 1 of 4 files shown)


oyamad:
Revised after review discussion — two changes, and rebased on `master` (#394):

**The CDF cache on the `MarkovChain` instance is removed.** The lazily initialized cache field made the first `simulate` call mutate the chain, so concurrent simulation on a shared instance — the natural threaded Monte Carlo pattern — became a data race (undefined behavior under Julia's memory model, even if benign on current hardware). A thread-safe version needs atomic lazy initialization, which deserves its own design discussion rather than a decision under merge pressure: opened #396 with the implementation sketch and the alternatives.

**Fixed a sparse index-type bug caught in external review.** The sparse sampler returned states of the matrix's index type `Ti`; for `Ti != Int` (e.g. `SparseMatrixCSC{Float64,Int32}`), the simulation iterator's `Tuple{Int,Int}` state annotation rejected the second step with a `MethodError`. The result is now converted to `Int` at the boundary, with a regression test.

## Benchmarks after the revision

Minimum times against the `master` baseline (9dcb458), both measured on the same machine — note this is a **different machine** from the one behind the numbers in the PR description, so the two tables are not comparable:

| Benchmark | `master` | This PR |
|:--|--:|--:|
| `simulate/sparse_n1000_k4_ts10000` | 2.03 ms | 146 μs |
| `stationary_distributions/dense_n200` | 1.83 ms | 853 μs |
| `simulate/dense_n1000_ts100` | 1.07 ms | 905 μs |
| `simulate/dense_n100_ts10000` | 306 μs | 269 μs |
| `simulate!/dense_n100_10000x10` | 2.98 ms | 2.63 ms |

The dedicated sparse sampler (~14x), the CDF-search overflow guard, the contiguous per-call dense CDFs, and the `recurrent_classes` shortcut all remain. What moved to #396 is the across-call amortization for repeated short simulations of a large dense chain (`simulate/dense_n1000_ts100`, 3.6 μs warm in the initial revision); within a single call, `simulate!` already amortizes the CDF construction across the columns of its output matrix.

On the `eltype` review comment: it is moot now that `MarkovChain` is back to three type parameters — though for the record, the method did dispatch as written: `MarkovChain{T,TM,TV}` applied to a four-parameter type is the partially applied `UnionAll` `MarkovChain{T,TM,TV,TS} where TS`, which matches every instance (`eltype` returned the correct state value type, and `simulate` allocated a correctly typed vector).

